PR #20, opened 2020-03-31
feat: [CBT-6 helper] Exposing Retry._deadline as a property

mf2199:
The suggested way of fixing cbt-6 involves propagating the value of a private member of the Retry class down to gRPC wrapper. The proposed changes are necessary to avoid the associated linting errors.

Fixes [#6](https://github.com/googleapis/python-bigtable/issues/6)


mf2199:
@kolea2 Friendly ping.

Issue #23, opened 2020-04-09
[AsyncIO] Overview of gRPC AsyncIO integration


lidizheng:
AsyncIO introduces `async`/`await` keywords. In order to plumb through the asynchronous-ness of functions, the library has to expose `async` functions on its surface. So, while integrating gRPC AsyncIO, we might need to instrument following classes in order to make gRPC AsyncIO work.

Also, since `python-api-core` supports all Python versions, we can't use any `async`/`await` in existing modules. If needed, the import of new AsyncIO modules will be protected by explicit version check.

A all-in-one draft PR can be found: https://github.com/googleapis/python-api-core/pull/22.

- [x] google.api_core.async_future
- [x] google.api_core.gapic_v1.config_async
- [x] google.api_core.gapic_v1.method_async
- [x] google.api_core.operations_v1.operations_async_client
- [x] google.api_core.grpc_helpers_async
- [x] google.api_core.operation_async
- [x] google.api_core.retry_async

Unit tests included for all new modules under tests/asyncio.

For each module, I will create separate issues and PRs later.

lidizheng:
With all components released, this issue should be closed.

PR #28, opened 2020-05-18
feat: AsyncIO Integration [Part 2]

lidizheng:
Children PR of https://github.com/googleapis/python-api-core/pull/26.

This PR includes AsyncIO version of:
* Polling future
* Page iterator

The AsyncIO version of polling future still uses the same mechanism as the sync version. The AsyncIO polling future tries to update its own state whenever the application want to access information or perform actions.

For page iterator, it has similar interface design as sync version. But instead of fulfilling normal generator protocol, it is based on the async generator.

Related #23

### DIFF
--- a/docs/operation.rst
+++ b/docs/operation.rst
@@ -4,3 +4,10 @@ Long-Running Operations
 .. automodule:: google.api_core.operation
   :members:
   :show-inheritance:
+
+Long-Running Operations in AsyncIO
+--------------
+
+.. automodule:: google.api_core.operation_async
+  :members:
+  :show-inheritance:
--- a/docs/operation.rst
+++ b/docs/operation.rst
@@ -6,7 +6,7 @@ Long-Running Operations
   :show-inheritance:
 
 Long-Running Operations in AsyncIO
---------------
+-------------------------------------
 
 .. automodule:: google.api_core.operation_async
   :members:
--- a/google/api_core/future/async_future.py
+++ b/google/api_core/future/async_future.py
@@ -1,0 +1,157 @@
+# Copyright 2020, Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""AsyncIO implementation of the abstract base Future class."""
+
+import asyncio
+
+from google.api_core import exceptions
+from google.api_core import retry
+from google.api_core import retry_async
+from google.api_core.future import base
+
+
+class _OperationNotComplete(Exception):
+    """Private exception used for polling via retry."""
+    pass
+
+
+RETRY_PREDICATE = retry.if_exception_type(
+    _OperationNotComplete,
+    exceptions.TooManyRequests,
+    exceptions.InternalServerError,
+    exceptions.BadGateway,
+)
+DEFAULT_RETRY = retry_async.AsyncRetry(predicate=RETRY_PREDICATE)
+
+
+class AsyncFuture(base.Future):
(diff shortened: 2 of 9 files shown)


google/api_core/operation_async.py
```diff
@@ -0,0 +1,217 @@
...
+    def _set_result_from_operation(self):
+        """Set the result or exception from the operation if it is complete."""
+        # This must be done in a lock to prevent the async_future thread
+        # and main thread from both executing the completion logic
+        # at the same time.
+        with self._completion_lock:
+            # If the operation isn't complete or if the result has already been
+            # set, do not call set_result/set_exception again.
+            # Note: self._result_set is set to True in set_result and
+            # set_exception, in case those methods are invoked directly.
```
busunkim96, line 124: Stale comment?

google/api_core/operation_async.py
```diff
@@ -0,0 +1,217 @@
...
+
+    operation = my_api_client.long_running_method()
+
+    def my_callback(future):
+        result = future.result()
+
+    operation.add_done_callback(my_callback)
+
+"""
+
```
busunkim96, line 38: Should these snippets use `await`?

google/api_core/future/async_future.py
```diff
@@ -0,0 +1,157 @@
...
+        If the operation is completed, the callback will be scheduled onto the
+        event loop. Otherwise, the callback will be stored and invoked when the
+        future is done.
+
+        Args:
+            fn (Callable[Future]): The callback to execute when the operation
+                is complete.
+        """
+        if self._background_task is None:
+            self._background_task = asyncio.get_event_loop().create_task(self._blocking_poll())
```
busunkim96, line 148: If I'm reading [this page](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) correctly, `asyncio.create_task` is only available in 3.7+?

docs/operation.rst
```diff
@@ -4,3 +4,10 @@ Long-Running Operations
 .. automodule:: google.api_core.operation
   :members:
   :show-inheritance:
+
+LRO in AsyncIO
```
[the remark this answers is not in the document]
lidizheng: Done.

google/api_core/future/async_future.py
lidizheng: Yes. `asyncio.create_task()` is 3.7+, but [`loop.create_task()`](https://docs.python.org/3.8/library/asyncio-eventloop.html#asyncio.loop.create_task) is 3.5+.

google/api_core/operation_async.py
lidizheng: Good catch! I missed this docstring. Updated to:
```
"""AsyncIO futures for long-running operations returned from Google Cloud APIs.

These futures can be used to await for the result of a long-running operation
using :meth:`AsyncOperation.result`:


.. code-block:: python

    operation = my_api_client.long_running_method()
    result = await operation.result()

Or asynchronously using callbacks and :meth:`Operation.add_done_callback`:

.. code-block:: python

    operation = my_api_client.long_running_method()

    def my_callback(future):
        result = await future.result()

    operation.add_done_callback(my_callback)

"""
```

google/api_core/operation_async.py
lidizheng: This note no longer apply to AsyncIO, removed.

docs/operation.rst
```diff
@@ -4,3 +4,10 @@ Long-Running Operations
 .. automodule:: google.api_core.operation
   :members:
   :show-inheritance:
+
+Long-Running Operations in AsyncIO
+--------------
```
busunkim96, line 9: (To fix the docs build)
```suggestion
-------------------------------------
```

lidizheng: One of reasons why Markdown is more popular than RST...

google/api_core/page_iterator_async.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        if self.page_number == 0:
+            return True
+
+        if self.max_results is not None:
+            if self.num_results >= self.max_results:
+                return False
+
+        # Note: intentionally a falsy check instead of a None check. The RPC
+        # can return an empty string indicating no more pages.
```
software-dov, line 277: Nit: move this comment to right above the 
```python
if self.max_results is not None
```

lidizheng:
Thanks for reviewing 👍

PR #36, opened 2020-06-03
feat(client_options): add new options 'quota_project' and 'credentialsAdd 'credentials_file' and 'quota_project' client options

software-dov:
Prerequisite for https://github.com/googleapis/gapic-generator-python/issues/432 and https://github.com/googleapis/gapic-generator-python/issues/433

### DIFF
--- a/google/api_core/client_options.py
+++ b/google/api_core/client_options.py
@@ -62,6 +62,9 @@ class ClientOptions(object):
             callback which returns client certificate file path, encrypted private
             key file path, and the passphrase bytes.``client_cert_source`` and
             ``client_encrypted_cert_source`` are mutually exclusive.
+        quota_project (str): An optional project name that a client's
+            quota belongs to.
+        credentials_file (str): An optional path to a file storing credentials.
 
     Raises:
         ValueError: If both ``client_cert_source`` and ``client_encrypted_cert_source``
@@ -73,6 +76,8 @@ class ClientOptions(object):
         api_endpoint=None,
         client_cert_source=None,
         client_encrypted_cert_source=None,
+        quota_project=None,
+        credentials_file=None,
     ):
         if client_cert_source and client_encrypted_cert_source:
             raise ValueError(
@@ -81,6 +86,8 @@ class ClientOptions(object):
         self.api_endpoint = api_endpoint
         self.client_cert_source = client_cert_source
         self.client_encrypted_cert_source = client_encrypted_cert_source
+        self.quota_project = quota_project
+        self.credentials_file = credentials_file
 
     def __repr__(self):
         return "ClientOptions: " + repr(self.__dict__)
--- a/google/api_core/client_options.py
+++ b/google/api_core/client_options.py
@@ -101,7 +101,7 @@ def from_dict(options):
     """Construct a client options object from a dictionary.
 
     Args:
-        options (dict[str, Any]): A dictionary with client options.
+        options (Dict[str, Any]): A dictionary with client options.
             See the docstring for ClientOptions for details on valid arguments.
     """
 
--- a/google/api_core/client_options.py
+++ b/google/api_core/client_options.py
@@ -63,7 +63,7 @@ class ClientOptions(object):
             A callback which returns client certificate file path, encrypted
             private key file path, and the passphrase bytes.``client_cert_source``
             and ``client_encrypted_cert_source`` are mutually exclusive.
-        quota_project (Optional[str]): A project name that a client's
+        quota_project_id (Optional[str]): A project name that a client's
             quota belongs to.
         credentials_file (Optional[str]): A path to a file storing credentials.
         scopes (Optional[Sequence[str]]): OAuth access token override scopes.
@@ -78,7 +78,7 @@ class ClientOptions(object):
         api_endpoint=None,
         client_cert_source=None,
         client_encrypted_cert_source=None,
-        quota_project=None,
+        quota_project_id=None,
         credentials_file=None,
         scopes=None,
     ):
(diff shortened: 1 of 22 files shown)


busunkim96:
Please hold off merging. Checking if we want quota project to be named `quota_project_id` or `quota_project`.

google/api_core/client_options.py
```diff
@@ -62,6 +62,9 @@ class ClientOptions(object):
             callback which returns client certificate file path, encrypted private
             key file path, and the passphrase bytes.``client_cert_source`` and
             ``client_encrypted_cert_source`` are mutually exclusive.
+        quota_project (str): An optional project name that a client's
+            quota belongs to.
+        credentials_file (str): An optional path to a file storing credentials.
```
plamut, line 67: (nit)
Optional arguments are usually annotated as `Optional[str]` (at least in PubSub and BigQuery libs that I know).

tests/unit/test_client_options.py
```diff
@@ -61,12 +66,18 @@ def test_constructor_with_both_cert_sources():
 
 def test_from_dict():
     options = client_options.from_dict(
-        {"api_endpoint": "foo.googleapis.com", "client_cert_source": get_client_cert}
+        {
+            "api_endpoint": "foo.googleapis.com",
+            "client_cert_source": get_client_cert,
+            "quota_project": "quote-proj",
+            "credentials_file": "path/to/credentials.json",
+        }
```
plamut, line 74: This might be unrelated, but I noticed the following lines in `from_dict()` method:
```py
if hasattr(client_options, key):
   ...

raise ValueError("ClientOptions does not accept an option '" + key + "'")
```
If `key` is not a string for some reason, both will result in a less pretty `TypeError`. Is this something we want to address?

(the docstring for the `options` argument is not specific on the types of acceptable the dict keys)

google/api_core/client_options.py
busunkim96: +1 We should probably update the args to use this style

tests/unit/test_client_options.py
software-dov: My vote is for making the docstring better instead of trying to field any random garbage that gets fed in.
The tricky part is that while the keys are always strings, the values are potentially arbitrary.
Does
```python
options (dict[str, Any]): ... See the docstring for ClientOptions for details on valid arguments.
```
fit the bill?

google/api_core/client_options.py
software-dov: Done

google/api_core/client_options.py
```diff
@@ -53,15 +53,19 @@ class ClientOptions(object):
...
-            callback which returns client certificate file path, encrypted private
-            key file path, and the passphrase bytes.``client_cert_source`` and
-            ``client_encrypted_cert_source`` are mutually exclusive.
+        client_encrypted_cert_source (Optional[Callable[[], (str, str, bytes)]]):
+            A callback which returns client certificate file path, encrypted
+            private key file path, and the passphrase bytes.``client_cert_source``
+            and ``client_encrypted_cert_source`` are mutually exclusive.
+        quota_project (Optional[str]): A project name that a client's
+            quota belongs to.
+        credentials_file (Optional[str]): A path to a file storing credentials.
```
busunkim96, line 68: Was the scope option left out intentionally? (https://github.com/googleapis/gapic-generator-python/issues/433)

tests/unit/test_client_options.py
plamut: This looks good, yes, thanks!

google/api_core/client_options.py
software-dov: Nope, not intentional, I just didn't see `scope`.

software-dov:
Who's involved in the `quota_project_id` vs `quota_project` decision?

busunkim96:
> Who's involved in the `quota_project_id` vs `quota_project` decision?

@broady @shinfan Does `quota_project_id` vs `quota_project` matter?

The Python auth lib uses `quota_project_id`, so I'd prefer to be consistent.

tests/unit/test_client_options.py
```diff
@@ -28,11 +28,18 @@ def get_client_encrypted_cert():
 def test_constructor():
 
     options = client_options.ClientOptions(
-        api_endpoint="foo.googleapis.com", client_cert_source=get_client_cert
+        api_endpoint="foo.googleapis.com",
+        client_cert_source=get_client_cert,
+        quota_project="quote-proj",
+        credentials_file="path/to/credentials.json",
+        scopes=["..", "..."]
```
busunkim96, line 35: Sorry, I was being lazy when I opened that issue and didn't find proper scopes. 😅 

Could you choose one or two from [this list](https://developers.google.com/identity/protocols/oauth2/scopes) and use them for these tests?

google/api_core/client_options.py
```diff
@@ -90,7 +101,8 @@ def from_dict(options):
     """Construct a client options object from a dictionary.
 
     Args:
-        options (dict): A dictionary with client options.
+        options (dict[str, Any]): A dictionary with client options.
```
busunkim96, line 104:
```suggestion
        options (Dict[str, Any]): A dictionary with client options.
```

tests/unit/test_client_options.py
software-dov: Done

shinfan:
quota_project_id sounds good tome

software-dov:
I'll wait until the beginning of tomorrow for a response from @broady or @shinfan . If I don't hear from either, I'll change to `quota_project_id` for assumed consistency.

PR #50, opened 2020-06-15
feat: allow credentials files to be passed for channel creation

busunkim96:
Allow credentials files to be passed to `create_channel`.

For https://github.com/googleapis/gapic-generator-python/pull/432

While it's possible for the library transport to create a credentials object and pass it to `create_channel`, I think it will be easier in the long run for the auth logic to remain in `_create_composite_credentials`.


google/api_core/grpc_helpers.py
```diff
@@ -218,7 +227,7 @@ def _create_composite_credentials(credentials=None, scopes=None, ssl_credentials
     )
 
 
-def create_channel(target, credentials=None, scopes=None, ssl_credentials=None, **kwargs):
+def create_channel(target, credentials=None, scopes=None, ssl_credentials=None, credentials_file=None, **kwargs):
```
busunkim96, line 230: I added `credentials_file` to the end of the list to avoid breaking anyone who is using positional arguments.

tests/asyncio/test_grpc_helpers_async.py
```diff
@@ -317,6 +317,19 @@ def test_create_channel_implicit_with_scopes(
...
+    target = "example:443"
+
+    with pytest.raises(ValueError) as excinfo:
+        grpc_helpers_async.create_channel(
+            target,
+            credentials_file="credentials.json",
+            credentials=mock.sentinel.credentials
+        )
+
+    assert "mutually exclusive" in str(excinfo.value)
```
software-dov, line 330: I'm not a huge fan of checking for explicit strings in exception messages. If it's easy and within the style guides of the repo, I like to make one-off exception types and check for the specific exception type within tests, something like
```python
class DuplicateCredentialArgs(Exception):
   pass
...
with pytest.raises(DuplicateCredentialArgs) as exc:
   ...
```

PR #81, opened 2020-10-02
test: test lower bounds

busunkim96:
Noxfile changes are the same as https://github.com/googleapis/python-bigquery/pull/263/.

Unit tests sessions pull dependency versions from a `constraints.txt` file. See [constraints files docs](https://pip.pypa.io/en/stable/user_guide/#constraints-files). The constraints file is populated for 2.7 and 3.6 and empty for remaining versions. Versions in the constraints files are the minimum version declared in the `setup.py` for each dependency. 

I also added the new [2020 pip resolver](https://discuss.python.org/t/announcement-pip-20-2-release/4863). It raised one warning on `setuptools` lower bounds as it is different from what is declared in `google-auth`. I'm not sure why that dependency is pinned here. It may be preferable to remove it.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -27,7 +27,7 @@ _MINIMAL_ASYNCIO_SUPPORT_PYTHON_VERSION = [3, 6]
 
 
 def _greater_or_equal_than_36(version_string):
-    tokens = version_string.split('.')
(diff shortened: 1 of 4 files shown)


busunkim96:
Obsolete, constraints files added in #122

PR #86, opened 2020-10-06
fix:  map LRO errors to library exception types

### LINKED ISSUE #15: Does not set status codes on operation errors
Raised in a [Cloud Spanner issue](https://github.com/googleapis/python-spanner/issues/38), the returned `GoogleAPICallError` has a `code` and `grpc_status_code` of `None`.

The issue seems to be at:
https://github.com/googleapis/python-api-core/blob/master/google/api_core/operation.py#L135

The `GoogleAPICallError` constructor does not set the code from the `errors` argument so the error code contained within it is being lost.

busunkim96:
Fixes #15  🦕

Errors raised by long running operations are currently always type `GoogleAPICallError`. Use the status code to create a more specific exception type.

### DIFF
--- a/google/api_core/exceptions.py
+++ b/google/api_core/exceptions.py
@@ -26,6 +26,7 @@ from six.moves import http_client
 
 try:
     import grpc
+
 except ImportError:  # pragma: NO COVER
     grpc = None
 
@@ -33,6 +34,14 @@ except ImportError:  # pragma: NO COVER
 # Populated by _APICallErrorMeta
 _HTTP_CODE_TO_EXCEPTION = {}
 _GRPC_CODE_TO_EXCEPTION = {}
+
+# Additional lookup table to map integer status codes to grpc status code
+# grpc does not currently support initializing enums from ints
+# i.e., grpc.StatusCode(5) raises an error
+if grpc is not None:
+    _INT_TO_GRPC_CODE = {}
+    for x in grpc.StatusCode:
+        _INT_TO_GRPC_CODE[x.value[0]] = x
 
 
 class GoogleAPIError(Exception):
@@ -432,7 +441,7 @@ def from_grpc_status(status_code, message, **kwargs):
     """Create a :class:`GoogleAPICallError` from a :class:`grpc.StatusCode`.
 
     Args:
-        status_code (grpc.StatusCode): The gRPC status code.
+        status_code (Union[grpc.StatusCode, int]): The gRPC status code.
         message (str): The exception message.
         kwargs: Additional arguments passed to the :class:`GoogleAPICallError`
             constructor.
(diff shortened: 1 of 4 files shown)


google/api_core/exceptions.py
```diff
@@ -441,6 +450,10 @@ def from_grpc_status(status_code, message, **kwargs):
         GoogleAPICallError: An instance of the appropriate subclass of
             :class:`GoogleAPICallError`.
     """
+
+    if isinstance(status_code, int):
+        status_code = _INT_TO_GRPC_CODE[status_code]
```
crwilcox, line 455: `_INT_TO_GRPC_CODE` only exists `if grpc`. Is this a problem? Also, do we want to get a with a default value (perhaps just return status code again?) in case the value isn't present?

google/api_core/exceptions.py
```diff
@@ -34,6 +35,14 @@
 _HTTP_CODE_TO_EXCEPTION = {}
 _GRPC_CODE_TO_EXCEPTION = {}
 
+# Additional lookup table to map integer status codes to grpc status code
+# grpc does not currently support initializing enums from ints
+# i.e., grpc.StatusCode(5) raises an error
+if grpc is not None:  # pragma: no branch
+    _INT_TO_GRPC_CODE = {}
```
tseaver, line 42: Move this up to unconditional module scope? (to address @crwilcox comment below).

google/api_core/exceptions.py
tseaver, line 455:
```suggestion
        status_code = _INT_TO_GRPC_CODE.get(status_code, status_code)
```

google/api_core/exceptions.py
```diff
@@ -34,6 +35,14 @@
...
+    for x in grpc.StatusCode:
+        _INT_TO_GRPC_CODE[x.value[0]] = x
+
```
tseaver, line 45:
```suggestion
# Additional lookup table to map integer status codes to grpc status code
# grpc does not currently support initializing enums from ints
# i.e., grpc.StatusCode(5) raises an error
_INT_TO_GRPC_CODE = {}
if grpc is not None:  # pragma: no branch
    for x in grpc.StatusCode:
        _INT_TO_GRPC_CODE[x.value[0]] = x

```

tseaver:
@googlebot I fixed it.

google/api_core/exceptions.py
busunkim96: Accepted @tseaver's suggestions (`_INT_TO_GRPC_CODE` always exists and  default value is `status_code`)

tseaver:
@googlebot I fixed it.

tseaver:
@busunkim96 I have no idea what the CLA bot is on about.

tseaver (approved):

PR #97, opened 2020-10-20
feat: add LogicalCallTimeout strategy

noahdietz:

### DIFF
--- a/google/api_core/timeout.py
+++ b/google/api_core/timeout.py
@@ -63,7 +63,8 @@ from google.api_core import general_helpers
 _DEFAULT_INITIAL_TIMEOUT = 5.0  # seconds
 _DEFAULT_MAXIMUM_TIMEOUT = 30.0  # seconds
 _DEFAULT_TIMEOUT_MULTIPLIER = 2.0
-_DEFAULT_OVERALL_TIMEOUT = 60.0  # seconds
+# Same as retry._DEFAULT_DEADLINE
+_DEFAULT_OVERALL_TIMEOUT = 60.0 * 2.0  # seconds
 # If specified, must be in seconds. If none, deadline is not used in the
 # timeout calculation.
 _DEFAULT_DEADLINE = None
--- a/google/api_core/timeout.py
+++ b/google/api_core/timeout.py
@@ -258,6 +258,12 @@ class LogicalCallTimeout(object):
 
     .. code-block:: python
 
+        logical_timeout = LogicalCallTimeout(100)
+
+        @logical_timeout
+        def make_api_call(request):
+            ...
+
     Args:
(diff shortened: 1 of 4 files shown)


google/api_core/timeout.py
```diff
@@ -222,3 +224,73 @@ def __str__(self):
...
+@six.python_2_unicode_compatible
+class LogicalCallTimeout(object):
+    """A decorator that adds a logical call timeout argument.
+
+    If called multiple times, each time this decorator will calculate a new
+    timeout parameter based on the time remaining in the overall timeout.
+
+    For example
+
+    .. code-block:: python
```
tseaver, line 259: This block needs to be populated, or it will break Sphinx.  E.g.:

```python
   logical_timeout = LogicalCallTimeout(100)

   @logical_timeout
   def make_api_call(foo, bar=None):
       ...
```

tseaver (commented):
I don't believe that this implementation can work in its current form:  see `google.api_core.retry.retry_target` for something which appears to do what you want here.

google/api_core/timeout.py
noahdietz: Ah ok, yeah this was a copy paste misstep from `ExponentialTimeout` declared above - it is also missing code block contents. I will add something.

noahdietz:
Hi @tseaver thanks for taking a look at this so long ago. I was on leave after opening it and I'm just coming back around to it. I'm not super familiar with python so your guidance is appreciated. 

> I don't believe that this implementation can work in its current form: see `google.api_core.retry.retry_target` for something which appears to do what you want here.

Can you be a little more specific? I was following the `ConstantTimeout` and `ExponentialTimeout` implementations as I felt this functionality should live in the same place considering it is intended to do something similar. I admit, it was mostly my poking around and guessing.

google/api_core/timeout.py
noahdietz: Done.

tseaver (changes requested):
Can we work on documenting why a user would chose to use the new timeout type, rather than configuring an `ExponentialTimeout` with the same `deadline`?

I *believe* from staring at the implementation that the goal is to bound the cumulative time for all back-off attempts to a hard maximum, but the `_exponential_timeout_generator` function [already does that](https://github.com/googleapis/python-api-core/blob/bdbf889210b709d7c1945f2160bcba9161b4dd2e/google/api_core/timeout.py#L134-L135).

If we *do* have a usecase for it, can we rename the class / helpers to indicate something more descriptive than "logical"?  And we likely need to document the interactions between the new timeout class and the exponential backoff in the sleep generator of the retry object.

busunkim96:
Closing as this PR has gone stale.

PR #98, opened 2020-10-22
chore: add a sync-repo-settings file

software-dov:


.github/sync-repo-settings.yaml
```diff
@@ -0,0 +1,16 @@
...
+  requiredApprovingReviewCount: 1
+  requiresCodeOwnerReviews: true
+  requiresStrictStatusChecks: true
+permissionRules:
+  - team: actools-python
+    permission: admin
+  - team: actools
+    permission: admin
+  - team: yoshi-python
+    permission: push
```
busunkim96, line 16: @JustinBeckwith Do these permission rules get combined with the default python rule or should we explicitly call out all the groups that need access?

JustinBeckwith: They are combined.  This is always additive.

busunkim96 (approved):

PR #134, opened 2021-01-20
feat: allow default_host and default_scopes to be passed to create_channel

busunkim96:
Add `default_host` and `default_scopes` parameters to `create_channel` so self-signed JWTs can be used. See https://github.com/googleapis/google-auth-library-python/pull/665 for the change in the auth library and https://github.com/googleapis/python-translate/pull/107 for the changes in the GAPIC library.

### DIFF
--- a/google/api_core/grpc_helpers.py
+++ b/google/api_core/grpc_helpers.py
@@ -17,6 +17,8 @@
 import collections
 
 import grpc
+from packaging import version
+import pkg_resources
 import six
 
 from google.api_core import exceptions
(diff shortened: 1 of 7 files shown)


google/api_core/grpc_helpers.py
```diff
@@ -210,21 +216,59 @@ def _create_composite_credentials(
...
+                scopes=scopes,
+                default_scopes=default_scopes
+            )
+        # google-auth < x.x.x does not have `default_scopes`
+        # TODO: remove this try/except once google-auth >= x.x.x is required
+        except TypeError:
+            credentials, _ = google.auth.load_credentials_from_file(
+                credentials_file,
+                scopes=scopes or default_scopes,
+            )
```
busunkim96, line 231: `x.x.x` as a place-holder. It will be updated with the real version once google-auth has been released.

We'll be able to test both branches for coverage since unit-3.6 will install a version *without* `default_scopes` and the other unit test runs will install versions *with* `default_scopes`. 

When this is merged, an issue will be created to track removing these try/excepts.

CC @tswast

arithmetic1728 (approved):

noxfile.py
```diff
@@ -50,6 +50,9 @@ def default(session):
     session.install("mock", "pytest", "pytest-cov", "grpcio >= 1.0.2")
     session.install("-e", ".", "-c", constraints_path)
 
+    # REMOVE ME: Temporarily install google-auth from a branch
+    session.install("-e", "git+https://github.com/googleapis/google-auth-library-python.git@self-signed-jwt#egg=google-auth")
```
arithmetic1728, line 54: `master` branch?

google/api_core/grpc_helpers.py
tswast: Thank you!

tswast: You might want to actually parse the versions, like we do in pandas-gbq. https://github.com/pydata/pandas-gbq/blob/81ab6b8f01c5a2cd1033e687a3d8bb7fe3f43c98/pandas_gbq/gbq.py#L61-L63

There are several options for this, the pandas-gbq library uses an old package from distutils/setuptools. It may be more reliable to use this package: https://packaging.pypa.io/en/latest/version.html#packaging.version.parse

busunkim96: Ah that looks like it will be cleaner. Will do!

busunkim96:
Reviewers, PTAL. Thank you!

arithmetic1728 (approved):
LGTM

tswast (approved):

PR #155, opened 2021-03-22
fix: save empty IAM policy bindings

### LINKED ISSUE #154: google.api_core.iam.Policy.__getitem__ does not correctly save empty bindings
We recently tried to upgrade a tool from 1.15.0 to 1.26.1 and found it breaks some code that handles IAM policies.  The commit that introduces the bug was released in 1.16.0: https://github.com/googleapis/python-api-core/commit/fd47fda5e3f5eca63522c8d81cffa22bc2a29ab6#diff-7cc73ea72342c139ff54060be9ff25b2f792f9225e0cc0f501dca9dbed9c4741 - 

The new `__getitem__` implementation returns a new empty `set()` for roles not in the current policy.  But it doesn't save that set in the bindings.  So if the user manipulates it, the policy isn't actually updated.  That breaks code written like this:
```python
policy = resource.get_iam_policy()
policy['roles/storage.objectAdmin'].add(principal)
bucket.set_iam_policy(policy)
```
This worked fine on v1.15.0 because of the use of `defaultdict`.  But now, this adds the principal to a set that's not used by the policy.

Something like the following (untested) patch should do the trick:
```diff
diff --git a/google/api_core/iam.py b/google/api_core/iam.py
index f130936..d650336 100644
--- a/google/api_core/iam.py
+++ b/google/api_core/iam.py
@@ -136,7 +136,9 @@ class Policy(collections_abc.MutableMapping):
         for b in self._bindings:
             if b["role"] == key:
                 return b["members"]
-        return set()
+
+        self[key] = set()
+        return self[key]

     def __setitem__(self, key, value):
         self.__check_version__()

```

busunkim96:
Fixes #154 

`__getitem__` currently returns an empty `set()` for items not in the bindings. If someone adds to the set returned by `__getitem__` the changes will not be reflected on the actual policy object.

### DIFF
--- a/tests/unit/test_iam.py
+++ b/tests/unit/test_iam.py
@@ -32,11 +32,11 @@ class TestPolicy:
         policy = self._make_one()
         assert policy.etag is None
         assert policy.version is None
+        assert len(policy) == 0
+        assert dict(policy) == {}
         assert policy.owners == empty
         assert policy.editors == empty
         assert policy.viewers == empty
-        assert len(policy) == 0
-        assert dict(policy) == {}
 
     def test_ctor_explicit(self):
         VERSION = 1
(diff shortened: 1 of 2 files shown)


tests/unit/test_iam.py
```diff
@@ -32,11 +32,11 @@ def test_ctor_defaults(self):
         policy = self._make_one()
         assert policy.etag is None
         assert policy.version is None
+        assert len(policy) == 0
+        assert dict(policy) == {}
```
busunkim96, line 36: Accessing `policy.owners`, `policy.editors` results in `__getitem__` being called, which adds empty entries to the dictionary. I moved these empty dict asserts up as a result.

An alternative is to keep them in their previous location and check that the three roles with empty bindings were added

parthea (approved):

PR #168, opened 2021-04-07
chore: remove reference to blacken in contrib

noahdietz:
Removes reference to the `blacken` `nox` session as it doesn't exist in the `noxfile.py`.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -98,11 +98,6 @@ On Debian/Ubuntu::
 ************
 Coding Style
 ************
-- We use the automatic code formatter ``black``. You can run it using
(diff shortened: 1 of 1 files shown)


plamut (commented):
(LGTM)

Thought - quite a few (most?) other libraries have the `blacken` session defined, would core also benefit if we added it to its noxfile?

busunkim96:
Yup it'd be nice to add `black` to this repo to. Opened https://github.com/googleapis/python-api-core/issues/169 to track that.

noahdietz:
I will close this then :) I was also kind of surprised because there is even a history of using `black` in this repo per the changelog. 🤷🏻‍♂️

PR #172, opened 2021-04-15
chore: prevent normalization of semver versioning

dandhlee:
When there is a patch version added to semver versioning, setuptools.setup(version) will normalize the versioning from -patch to .patch which is not correct SEMVER versioning. The added feature with setuptools.sic(version) will prevent this from happening.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ if "google.cloud" in packages:
 
 setuptools.setup(
     name=name,
-    version=version,
+    version=setuptools.sic(version),
     description=description,
     long_description=readme,
     author="Google LLC",
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,21 @@ import os
 
 import setuptools
 
+# Disable version normalization performed by setuptools.setup()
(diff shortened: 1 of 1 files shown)


parthea (dismissed):

setup.py
```diff
@@ -17,6 +17,21 @@
...
+except ImportError:
+    # Try the approach of replacing packaging.version.Version
+    sic = lambda v: v
+    try:
+        # setuptools >=39.0.0 uses packaging from setuptools.extern
+        from setuptools.extern import packaging
+    except ImportError:
+        # setuptools <39.0.0 uses packaging from pkg_resources.extern
+        from pkg_resources.extern import packaging
+    packaging.version.Version = packaging.version.LegacyVersion
```
parthea, line 34: Remove work around and pin to setuptools version here:
https://github.com/googleapis/python-api-core/blob/master/setup.py#L36

parthea: My mistake. This is needed for python 2 support.

parthea (approved):

PR #189, opened 2021-05-17
feat: Add support for `rest/` token in `x-goog-api-client` header

vam-google:
This is required for (DI)REGAPIC clients, like GCE client.

### DIFF
--- a/google/api_core/client_info.py
+++ b/google/api_core/client_info.py
@@ -54,6 +54,7 @@ class ClientInfo(object):
         user_agent (Optional[str]): Prefix to the user agent header. This is
             used to supply information such as application name or partner tool.
             Recommended format: ``application-or-tool-ID/major.minor.version``.
+        rest_version (Optional[str]): The requests library version.
     """
 
     def __init__(
@@ -64,6 +65,7 @@ class ClientInfo(object):
         gapic_version=None,
         client_library_version=None,
         user_agent=None,
+        rest_version=None,
     ):
         self.python_version = python_version
         self.grpc_version = grpc_version
@@ -71,6 +73,7 @@ class ClientInfo(object):
         self.gapic_version = gapic_version
         self.client_library_version = client_library_version
         self.user_agent = user_agent
+        self.rest_version = rest_version
 
     def to_user_agent(self):
         """Returns the user-agent string for this client info."""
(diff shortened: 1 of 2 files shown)


vam-google:
The test are failing because of missing `grpc` dependency, which is listed as optional (thought it is not really optional). It has been like this before (should not be related to this specific change).

google/api_core/client_info.py
```diff
@@ -64,13 +65,15 @@ def __init__(
...
         user_agent=None,
+        rest_version=None,
     ):
         self.python_version = python_version
         self.grpc_version = grpc_version
         self.api_core_version = api_core_version
         self.gapic_version = gapic_version
         self.client_library_version = client_library_version
         self.user_agent = user_agent
+        self.rest_version = rest_version
```
software-dov, line 76: Should there be a check that either `rest_version` XOR `grpc_version` is set?

software-dov (approved):

google/api_core/client_info.py
vam-google: Sounds reasonable to me, I was thinking about the same thing, but then decided that it would be safer to not add something like tha to a GA'ed library. If you feel strong about it, I can add it. Should I?

software-dov: I don't feel _strongly_ about it, but it seems like a reasonable thing for someone looking at the fields to expect.

busunkim96:
@vam-google It looks like lint is failing (everything else is passing)

```
nox > flake8 google tests
tests/unit/test_client_info.py:74:1: E302 expected 2 blank lines, found 1
```

```
nox > Session docfx was successful.
nox > Ran multiple sessions:
nox > * unit-2.7: success
nox > * unit-3.6: success
nox > * unit-3.7: success
nox > * unit-3.8: success
nox > * unit-3.9: success
nox > * unit_grpc_gcp-2.7: success
nox > * unit_grpc_gcp-3.6: success
nox > * unit_grpc_gcp-3.7: success
nox > * unit_grpc_gcp-3.8: success
nox > * unit_grpc_gcp-3.9: success
nox > * lint: failed
nox > * lint_setup_py: success
nox > * pytype: success
nox > * cover: success
nox > * docs: success
nox > * docfx: success
```

PR #30, opened 2020-05-21
feat: allow disabling response stream pre-fetch

plamut:
Closes #25.

This PR adds the ability to disable automatically pre-fetching the first item of a stream returned by `*-Stream` gRPC callables. This hook will be used in PubSub to fix the [stalled stream issue](https://github.com/googleapis/python-pubsub/issues/93), while also not affecting Firestore, since the default behavior is preserved.

I realize the fix is far from ideal, but it's the least ugly among the approaches I tried, e.g. somehow passing the flag through `ResumableBidiRpc` (it's a messy rabbit hole). On the PubSub side monkeypatching the generated SubscriberClient will be needed, but it's a (relatively) clean one-liner:
```patch
diff --git google/cloud/pubsub_v1/gapic/subscriber_client.py google/cloud/pubsub_v1/gapic/subscriber_client.py
index e98a686..1d6c058 100644
--- google/cloud/pubsub_v1/gapic/subscriber_client.py
+++ google/cloud/pubsub_v1/gapic/subscriber_client.py
@@ -1169,6 +1169,8 @@ class SubscriberClient(object):
                 default_timeout=self._method_configs["StreamingPull"].timeout,
                 client_info=self._client_info,
             )
+            # TODO: explain this monkeypatch!
+            self.transport.streaming_pull._prefetch_first_result_ = False
 
         return self._inner_api_calls["streaming_pull"](
             requests, retry=retry, timeout=timeout, metadata=metadata
```

If/when we merge this, we should also release it, and then we can add `!= 1.17.0` to the `google-api-core` version pin in PubSub.

### PR checklist
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-api-core/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

PR #205, opened 2021-06-10
feat: Bidi uses metadata from start_rpc

### LINKED ISSUE #202: `client_info` from `start_rpc` arg should be used when creating `BidiRpc`
https://github.com/googleapis/python-api-core/blob/9eaa7868164a7e98792de24d2be97f79fba22322/google/api_core/bidi.py#L246

GAPIC clients add `client_info` metadata to rpcs when the methods are wrapped. BidiRPC should use this start rpc's metadata if `metadata` is not passed to the constructor.

CC @jameslynnwu

busunkim96:
Allow Bidi to use metadata from start_rpc when no other metadata
is provided.

This allows `client_info` on wrapped client methods will be passed through to `BidiRpc`.

Fixes #202 🦕

Not expected to impact Firestore, as metadata is explicitly passed: https://github.com/googleapis/python-firestore/blob/e57258c51e4b4aa664cc927454056412756fc7ac/google/cloud/firestore_v1/watch.py#L220-L226

Pub/Sub does not seem to pass `metadata` and may be impacted: https://github.com/googleapis/python-pubsub/blob/e907f6e05f59f64a3b08df3304e92ec960997be6/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py#L523-L528

### DIFF
--- a/google/api_core/bidi.py
+++ b/google/api_core/bidi.py
@@ -240,7 +240,8 @@ class BidiRpc(object):
             yield. This is useful if an initial request is needed to start the
             stream.
         metadata (Sequence[Tuple(str, str)]): RPC metadata to include in
-            the request.
+            the request. If no metadata is provided, metadata from
+            start_rpc will be used.
     """
 
     def __init__(self, start_rpc, initial_request=None, metadata=None):
@@ -277,7 +278,11 @@ class BidiRpc(object):
         request_generator = _RequestQueueGenerator(
             self._request_queue, initial_request=self._initial_request
         )
-        call = self._start_rpc(iter(request_generator), metadata=self._rpc_metadata)
+        if self._rpc_metadata:
+            call = self._start_rpc(iter(request_generator), metadata=self._rpc_metadata)
+        # use metadata from self._start_rpc if no other metadata is specified
+        else:
+            call = self._start_rpc(iter(request_generator))
 
         request_generator.call = call
 
--- a/tests/unit/test_bidi.py
+++ b/tests/unit/test_bidi.py
@@ -202,12 +202,12 @@ class _CallAndFuture(grpc.Call, grpc.Future):
     pass
 
 
-def make_rpc():
+def make_rpc(metadata=None):
     """Makes a mock RPC used to test Bidi classes."""
     call = mock.create_autospec(_CallAndFuture, instance=True)
     rpc = mock.create_autospec(grpc.StreamStreamMultiCallable, instance=True)
 
-    def rpc_side_effect(request, metadata=None):
+    def rpc_side_effect(request, metadata=metadata):
         call.is_active.return_value = True
         call.request = request
         call.metadata = metadata
@@ -265,12 +265,13 @@ class TestBidiRpc(object):
         assert bidi_rpc.call.metadata == mock.sentinel.A
 
     def test_open(self):
-        rpc, call = make_rpc()
-        bidi_rpc = bidi.BidiRpc(rpc)
+        rpc, call = make_rpc(metadata=[(1, 2)])
+        bidi_rpc = bidi.BidiRpc(rpc, metadata=[(3, 4)])
 
         bidi_rpc.open()
 
         assert bidi_rpc.call == call
+        assert bidi_rpc.call.metadata == [(3, 4)]
         assert bidi_rpc.is_active
         call.add_done_callback.assert_called_once_with(bidi_rpc._on_call_done)
 
@@ -282,6 +283,14 @@ class TestBidiRpc(object):
 
         with pytest.raises(ValueError):
             bidi_rpc.open()
+
+    def test_open_use_start_rpc_metadata(self):
+        rpc, _ = make_rpc(metadata=[(1, 2)])
+        bidi_rpc = bidi.BidiRpc(rpc)
+
+        bidi_rpc.open()
+
+        assert bidi_rpc.call.metadata == [(1, 2)]
 
     def test_close(self):
         rpc, call = make_rpc()


google/api_core/bidi.py
```diff
@@ -277,7 +278,11 @@ def open(self):
         request_generator = _RequestQueueGenerator(
             self._request_queue, initial_request=self._initial_request
         )
-        call = self._start_rpc(iter(request_generator), metadata=self._rpc_metadata)
+        if self._rpc_metadata:
+            call = self._start_rpc(iter(request_generator), metadata=self._rpc_metadata)
+        # use metadata from self._start_rpc if no other metadata is specified
+        else:
+            call = self._start_rpc(iter(request_generator))
```
jameslynnwu, line 285: When I set up BidiRPC with `LoggingV2Transport().tail_log_entries`, that's the unwrapped method, not the wrapped method. Will this still get the metadata from the wrapped method?

tseaver (commented):
Overall, LGTM.  I'll let @plamut comment on possible Pub/sub interactions.

google/api_core/bidi.py
jameslynnwu: also, I tried passing in LoggingServiceV2Client().tail_log_entries and it didn't seem to work.

I'm actually not sure how pubsub passes in the wrapped method successfully.
https://github.com/googleapis/python-pubsub/blob/e907f6e05f59f64a3b08df3304e92ec960997be6/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py#L524

busunkim96: > When I set up BidiRPC with LoggingV2Transport().tail_log_entries, that's the unwrapped method, not the wrapped method. Will this still get the metadata from the wrapped method?

I believe the wrapped one has to be passed in (the wrapper adds the metadata to the call)

> also, I tried passing in LoggingServiceV2Client().tail_log_entries and it didn't seem to work.
>
> I'm actually not sure how pubsub passes in the wrapped method successfully.

Hmm I'll go back and poke at what is actually expected to be passed in to Bidi.

busunkim96:
Looking at Firestore vs. Pub/Sub, I see Firestore is passing the method in the [gRPC transport](https://github.com/googleapis/python-firestore/blob/e57258c51e4b4aa664cc927454056412756fc7ac/google/cloud/firestore_v1/services/firestore/transports/grpc.py#L554) and Pub/Sub is passing the method in the [GAPIC client](https://github.com/googleapis/python-pubsub/blob/d0d0b704642df8dee893d3f585aeb666e19696fb/google/pubsub_v1/services/subscriber/client.py#L1152).

I wonder if that contributed to the behavior difference between the two that requires [this workaround](https://github.com/googleapis/python-pubsub/blob/d0d0b704642df8dee893d3f585aeb666e19696fb/google/pubsub_v1/services/subscriber/client.py#L1190-L1193)? (Also see: https://github.com/googleapis/python-pubsub/issues/93#issuecomment-630762257 and https://github.com/googleapis/python-api-core/pull/30)

@jameslynnwu For it "not working" when `LoggingServiceV2Client().tail_log_entries` was passed did it hang, error, or something else?

jameslynnwu:
Great find!
When I pass in `LoggingServiceV2Client().tail_log_entries`, it doesn't respond and the request does not seem to be initiated.

However, if I use the same workaround, it works.
```
# Does not respond. Request doesn't seem to be initiated
client = LoggingServiceV2Client()
self.tail_stub = bidi.BidiRpc(client.logging.tail_log_entries)

# Works
client = LoggingServiceV2Client()
self.tail_stub = bidi.BidiRpc(client.logging.transport.tail_log_entries)

# Works
client = LoggingServiceV2Client()
client.transport._prefetch_first_result_ = False
self.tail_stub = bidi.BidiRpc(client.logging.tail_log_entries)
```

plamut:
The Pub/Sub client passes the `streaming_pull()` method when creating a [Bidi instance](https://github.com/googleapis/python-pubsub/blob/7597604b41fa3a1e9bf34addc35c8647dde007cc/google/cloud/pubsub_v1/subscriber/_protocol/streaming_pull_manager.py#L523-L529). This method comes from the auto-generated client and [invokes](https://github.com/googleapis/python-pubsub/blob/d0d0b704642df8dee893d3f585aeb666e19696fb/google/pubsub_v1/services/subscriber/client.py#L1152-L1203) the wrapped transport method.

The subscriber client does not pass any metadata on its own to `ResumableBidiRpc`, it just leaves it up to the (generated) transport to do the [wrapping](https://github.com/googleapis/python-pubsub/blob/d0d0b704642df8dee893d3f585aeb666e19696fb/google/pubsub_v1/services/subscriber/transports/base.py#L227-L244), which sets `client_info` on requests, among other things.

---
As for the changes in this PR - since no metadata is passed to `BidiRpc`, the default value `None` is used for it.

Before the PR, `BidiRpc.open()` called the following (sending `None` for metadata:
```py
call = self._start_rpc(iter(request_generator), metadata=self._rpc_metadata)
```

With the changes in this PR, however, that changes:
```py
if self._rpc_metadata:
    call = self._start_rpc(iter(request_generator), metadata=self._rpc_metadata)
# use metadata from self._start_rpc if no other metadata is specified
else:
    call = self._start_rpc(iter(request_generator))
```
Since metadata is `None`, this code now effectively becomes the following:
```py
 call = self._start_rpc(iter(request_generator))
```
No explicit `None` metadata is passed in, thus the generated `streaming_pull()` method now uses [its own default](https://github.com/googleapis/python-pubsub/blob/d0d0b704642df8dee893d3f585aeb666e19696fb/google/pubsub_v1/services/subscriber/client.py#L1158), which is an empty tuple.

When the (wrapped) method is eventually called, the [following logic](https://github.com/googleapis/python-api-core/blob/11032cf08ecc16dd252a6cda8b33b0b28ec4f4ba/google/api_core/gapic_v1/method.py#L135-L141) in `_GapicCallable` determines the metadata:
```py
if self._metadata is not None:
    metadata = kwargs.get("metadata", [])
    # Due to the nature of invocation, None should be treated the same
    # as not specified.
    if metadata is None:
        metadata = []
    metadata = list(metadata)
    ...
```

If I traced this correctly, there should be no difference between passing `None` and an empty tuple as a kwarg. Both get transformed to an empty list, and the original metadata from the method wrapping time is not modified.

jameslynnwu:
One more thing to note, though we probably not want to address all of this at once.

Passing in the `client.logging.tail_log_entries` vs `client.logging.transport.tail_log_entries` changes what is passed into the `should_recover` method of `ResumableBidiRPC` from either a [grpc.Call](https://grpc.github.io/grpc/python/grpc.html#grpc.Call) or a [google.api_core.exceptions.GoogleAPICallError](https://googleapis.dev/python/google-api-core/latest/exceptions.html#google.api_core.exceptions.GoogleAPICallError) to a [grpc.Call](https://grpc.github.io/grpc/python/grpc.html#grpc.Call), respectively. 

If we can ensure that only `client.logging.tail_log_entries` is passed it that would make `should_recover` more predictable. Alternatively, we should update the code to always pass the same result into `should_recover` regardless of if `client.logging.tail_log_entries` or `client.logging.transport.tail_log_entries` is passed in.

Also, it's interesting that `should_recover` is called twice from callbacks and from exception.

busunkim96:
@plamut Thank you for the detailed analysis. 🙏 

 I'm OOO the rest of the week so I'll come back and look at this next week.

busunkim96:
As @plamut explained above, this PR is a no-op, if you pass the wrapped method the metadata will get added (which is why Pub/Sub doesn't pass `metadata` explicitly to `start_rpc`).

I had a bit of trouble tracing this through but was able to confirm the expected headers in the gRPC logs:

```
export GRPC_TRACE=all
export GRPC_VERBOSITY=debug
```

```
I0625 21:17:38.217348925 2619277 chttp2_transport.cc:1363]   HTTP:0:HDR:CLI: :scheme: https
I0625 21:17:38.217356118 2619277 chttp2_transport.cc:1363]   HTTP:0:HDR:CLI: :method: POST
I0625 21:17:38.217363271 2619277 chttp2_transport.cc:1363]   HTTP:0:HDR:CLI: :authority: pubsub.googleapis.com:443
I0625 21:17:38.217370411 2619277 chttp2_transport.cc:1363]   HTTP:0:HDR:CLI: :path: /google.pubsub.v1.Subscriber/StreamingPull
I0625 21:17:38.217377702 2619277 chttp2_transport.cc:1363]   HTTP:0:HDR:CLI: x-goog-api-client: gl-python/3.8.3 grpc/1.38.0 gax/1.30.0 gccl/2.5.0
```

PR #212, opened 2021-06-15
fix!: drop support for Python 2.7 / 3.5

### LINKED ISSUE #210: Drop Python 2.7 support
This would be semver major (`google-api-core` 2.0.0).  Clients which cannot move forward (e.g, `google-cloud-storage` already pin `google.api_core < 2.0.0dev` either directly or transitively (via `google-cloud-core`).

Extra cleanups:

- ~~Drop `mock` wrapper (use `from unittest import mock` directly, replacing work in #208).~~  Can't remove it until we drop support for Python 3.6 / 3.7, as it provides the `mock.AsyncMock` helper we need.
- [x] Drop `six`.
- [x] Drop other Python 2.7-specific workarounds (`collections_abc`, etc.)
- [x] Drop `u"` prefixes for text strings.
- [x] Drop `pytz` (per #73).
- ~~Replace [workaround](https://github.com/googleapis/python-api-core/blob/11032cf08ecc16dd252a6cda8b33b0b28ec4f4ba/google/api_core/bidi.py#L640-L652) in `bidi.BackgroundConsumer._thread_main` with `Condition.wait_for`.~~ Moved to #211.

tseaver:
- Drop `six`.
- Drop other Python 2.7-specific workarounds (`collections_abc`, etc.)
- Drop `u"` prefixes for text strings.
- Drop `pytz` (per googleapis/python-api-core#73).

Release-As: 2.0.0b1

Closes googleapis/python-api-core#210.


README.rst
```diff
@@ -16,8 +16,4 @@ common helpers used by all Google API clients. For more information, see the
 
 Supported Python Versions
 -------------------------
-Python >= 3.5
-
-Deprecated Python Versions
---------------------------
-Python == 2.7. Python 2.7 support will be removed on January 1, 2020.
```
plamut, line 23: IIRC this section should be retained, if only to explain what the last Python 2.7 compatible version is.

BigQuery example: https://googleapis.dev/python/bigquery/latest/index.html#unsupported-python-versions
Pub/Sub docs example: https://googleapis.dev/python/pubsub/latest/index.html#deprecated-python-versions

google/api_core/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "1.30.0"
+__version__ = "2.0.0-b1"
```
plamut, line 15: Is this a new process, i.e. having a preliminary release instead of immediately switching to 2.0.0? Also, should this be beta or an RC?

FWIW, the approach itself does seem sensible to me.

google/api_core/future/base.py
```diff
@@ -16,11 +16,8 @@
 
 import abc
 
-import six
```
plamut, line 19: I noticed that we still have a `six` pin in `testing/constraints-3.6.txt`, let's remove that, too.

tests/unit/test_exceptions.py
```diff
@@ -138,15 +138,15 @@ def test_from_http_response_bad_json_content():
...
-            {"error": {"message": u"\u2019 message", "errors": ["1", "2"]}}
+            {"error": {"message": "\u2019 message", "errors": ["1", "2"]}}
         ).encode("utf-8")
     )
 
     exception = exceptions.from_http_response(response)
 
     assert isinstance(exception, exceptions.NotFound)
     assert exception.code == http.client.NOT_FOUND
-    assert exception.message == u"POST https://example.com/: \u2019 message"
```
plamut, line 149: I found a few more of these in `docs/conf.py`, let's remove them there, too.

plamut (commented):
Generally looks good, just spotted a few legacy bits that were not cleaned yet.

There are also a few remaining references to _"2.7"_ and _"Python 2"_ (comments, workarounds, etc.) that can be removed. And maybe the [workaround](https://github.com/googleapis/python-api-core/blob/11032cf08ecc16dd252a6cda8b33b0b28ec4f4ba/google/api_core/bidi.py#L640-L652) for the lack of `Condition.wait_for()` method in Python 2.7.

---

**Edit:** Oh, and we also use the `mock` shim in unit tests (installed in noxfile's default session), we can now switch to importing the standard library's `unittest.mock`.

**Edit 2:** We can also get rid of several `def next(self)` methods and replace them with `__next__(self)`.

google/api_core/version.py
tseaver: As a semver-major / breaking change, I thought it was reasonable to do a short beta.  I could be persuaded that it should be an RC, but I want folks to consider carefully whether they install it in production until we've gotten feedback that all is OK.

tests/unit/test_exceptions.py
tseaver: Templated file -- I'd rather fix that upstream in `synthtool`.

google/api_core/future/base.py
tseaver: 6e21b01

README.rst
tseaver: fe7d454

tseaver:
@plamut Thanks for the review

> There are also a few remaining references to "2.7" and "Python 2" (comments, workarounds, etc.) that can be removed.

I'll leave the bit in `docs/auth.rst` for now.  I'm working through a couple of others.

> And maybe the workaround for the lack of Condition.wait_for() method in Python 2.7.

I added #211 to track that change:  I didn't want to think that hard in the middle of a big janitorial cleanup.

> Edit: Oh, and we also use the `mock` shim in unit tests (installed in noxfile's default session), we can now switch
> to importing the standard library's unittest.mock.

We can't replace `import mock` wholesale with `from unittest import mock`, because the stdlib version doesn't have 'mock.AsyncMock` before Python 3.8.

> Edit 2: We can also get rid of several def next(self) methods and replace them with __next__(self).

4165633, ad7b40b, dbc234f.

tests/unit/test_exceptions.py
tseaver: https://github.com/googleapis/synthtool/pull/1119

plamut (approved):
LGTM, thanks for the additional cleanup.

It's totally fine to defer some of the less trivial (and thus risky) changes to a separate PR, esp. with issues to track them.

setup.py
```diff
@@ -35,8 +35,6 @@
     "requests >= 2.18.0, < 3.0.0dev",
     "setuptools >= 40.3.0",
     "packaging >= 14.3",
```
parthea, line 37: Can we remove packaging also? See #215

tseaver:
Marked as `do not merge` pending discussion over the versioning policy next week with @busunkim96, @crwilcox, et al.

owlbot.py
```diff
@@ -22,8 +22,15 @@
 # ----------------------------------------------------------------------------
 # Add templated files
 # ----------------------------------------------------------------------------
+excludes = [
+    "noxfile.py",  # pytype
+    "setup.cfg",  # pytype
+    ".flake8",  # flake8-import-order, layout
+    ".coveragerc",  # layout
+    "CONTRIBUTING.rst",  # no systests
```
parthea, line 30: Can we keep `CONTRIBUTING.rst` and just update the SHA in `.OwlBot.lock.yaml` to `b6169fc6a5207b11800a7c002d0c5c2bc6d82697185ca12e666f44031468cfcd`?

tseaver: There should be no "System Tests" section at all in `CONTRIBUTING.rst` for this repository -- it has none.  I don't know that the generator is configurable enough to allow that change, and I'd really like not to have to add the monster `s.replace` required to remove it.

parthea: Makes sense.

setup.py
tseaver: 72d175b

tseaver:
Ugh, OwlBot is restoring the `2.7` reference in the templated `scripts/readme-gen/templates/install_deps.tmpl.rst'.  See https://github.com/googleapis/synthtool/pull/1173.

PR #228, opened 2021-07-15
fix: strip trailing `_` from field mask paths

### LINKED ISSUE #227: Modified reserved word field names included in field masks.
I'm not sure whether this is an issue with this library, the [`generator`](https://github.com/googleapis/gapic-generator-python) or the [`field mask helper`](https://github.com/googleapis/python-api-core/blob/master/google/api_core/protobuf_helpers.py#L290), but it seems to be incorrect behavior that's blocking for Ads users, unless there's a workaround or other solution I'm not aware of:

#### Environment details

  - Programming language: `Python`
  - OS: `gLinux`
  - Language runtime version: `3.7.0`
  - Package version: `1.18.1`

#### Steps to reproduce

Assuming we have a protobuf definition like the following:

```
message HasType {
  string type = 1;
}
```

We'll have a generated Python class similar to:

```py
class HasType(proto.Message):
    type_ = proto.Field(proto.STRING, number=1,)
```

Where `type` has been modified to `type_`. 

Normally setting `type_` isn't a problem, the issue is that this modified field name gets leaked into field masks, which are revoked by the API because it doesn't match the protobuf definition. This prevents users from executing update requests where the `type` field on any object has been modified. 

Here's a short script to reproduce:

```py
from google.api_core import protobuf_helpers
import proto

class HasType(proto.Message):
    type_ = proto.Field(proto.STRING, number=1,)

ht = HasType()
ht.type_ = "has type"

fm = protobuf_helpers.field_mask(None, ht._pb)

print(fm.paths[0])
# >>> "type_"
```

busunkim96:

CC @benrkarl

Fixes #227🦕

### DIFF
--- a/google/api_core/protobuf_helpers.py
+++ b/google/api_core/protobuf_helpers.py
@@ -357,6 +357,13 @@ def _field_mask_helper(original, modified, current=""):
 
 
 def _get_path(current, name):
+    # gapic-generator-python appends underscores to field names
+    # that collide with python keywords.
+    # `_` is stripped away as it is not possible to
+    # natively define a field with a trailing underscore in protobuf.
+    # APIs will reject field masks if fields have trailing underscores.
+    # See https://github.com/googleapis/python-api-core/issues/227
+    name = name.rstrip("_")
     if not current:
         return name
     return "%s.%s" % (current, name)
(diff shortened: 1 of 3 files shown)


tseaver (approved):
Unfortunate that we have to add more Python2-specific cruft, just at this point.

busunkim96:
@software-dov Could you take a look? Thank you!

busunkim96:
@software-dov Friendly ping

software-dov (approved):

PR #234, opened 2021-07-22
fix: bump grpcio version to use stable aio API

lidizheng:
Related https://github.com/googleapis/python-api-core/issues/40.

gRPC AsyncIO API has been stable for more than a year now. Existing import of `grpc.experimental.aio` works, because it's an alias of `grpc.aio`. Still, it's nice to use the more correct version.

PR #246, opened 2021-08-06
chore: avoid duplicating pins of grpcio in noxfile

tseaver:
/cc @lidizheng 

Rely on the pins in 'setup.py' as the Source of Truth.

See https://github.com/googleapis/python-api-core/pull/234#pullrequestreview-724669326


busunkim96 (approved):
Eek I didn't realize this was the case. Thank you for fixing this!

PR #248, opened 2021-08-06
feat: add api_key to client options

arithmetic1728:


tests/unit/test_client_options.py
```diff
@@ -109,9 +140,10 @@ def test_from_dict_bad_argument():
 
 def test_repr():
     options = client_options.ClientOptions(api_endpoint="foo.googleapis.com")
+    print(repr(options))
```
parthea, line 143: If this is for debugging purposes, should we remove it?

parthea (approved):
Added one minor observation

busunkim96 (approved):

PR #258, opened 2021-08-20
fix: do not error on LROs with no response or error

### LINKED ISSUE #257: Do not throw error on empty LRO response
[operation.py](https://github.com/googleapis/python-api-core/blob/master/google/api_core/operation.py) throws an error if an LRO finishes but returns an empty response (the error message that it throws is "Unexpected state: Long-running operation had neither response nor error set"). However some APIs, such as Contact Center Insights, end LRO without setting either result or error (but set `done: true`). For example, please see [`ExportInsightsDataResponse`](https://cloud.google.com/contact-center/insights/docs/reference/rpc/google.cloud.contactcenterinsights.v1#exportinsightsdataresponse) which is an empty response that is returned after the `ExportInsightsData` LRO is complete.

We should update [operation.py](https://github.com/googleapis/python-api-core/blob/master/google/api_core/operation.py) to return an empty response instead of an error when it encounters an empty LRO. For reference, we found a similar issue in the NodeJS operation library and it was fixed [here](https://github.com/googleapis/gax-nodejs/pull/1087).

busunkim96:
> The response type should not be google.protobuf.Empty (except for Delete methods), unless it is certain that response data will never be needed. If response data might be added in the future, define an empty message for the RPC response and use that.
>

https://google.aip.dev/151 

Some APIs will set `done: true` but have an empty response and exception.
Set the result to an empty message of the expected type instead
of erroring.

Fixes #257 🦕


parthea (approved):

tseaver (approved):

google/api_core/operation.py
```diff
@@ -139,12 +139,12 @@ def _set_result_from_operation(self):
...
-                    "Unexpected state: Long-running operation had neither "
-                    "response nor error set."
-                )
-                self.set_exception(exception)
+            elif self._operation.done:
+                # Some APIs set `done: true`, with an empty response.
+                # Set the result to an empty message of the expected
+                # result type.
+                # https://google.aip.dev/151
+                self.set_result(self._result_type())
```
tseaver, line 147: Coverage thinks we need a final `else:` here (for the case that `self._operation.done` is False).

google/api_core/operation_async.py
```diff
@@ -135,12 +135,12 @@ def _set_result_from_operation(self):
...
-                    "Unexpected state: Long-running operation had neither "
-                    "response nor error set."
-                )
-                self.set_exception(exception)
+            elif self._operation.done:
+                # Some APIs set `done: true`, with an empty response.
+                # Set the result to an empty message of the expected
+                # result type.
+                # https://google.aip.dev/151
+                self.set_result(self._result_type())
```
tseaver, line 143: Coverage thinks we need a final `else:` here (for the case that `self._operation.done` is False).

tseaver (changes requested):

google/api_core/operation.py
busunkim96: 👍  I switched this back to `else:` as `self._operation.done` must already be `True` to get to this clause.

tseaver (approved):

PR #259, opened 2021-08-25
feat: add grpc transcoding + tests

yihjenku:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-api-core/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass

Fixes #<issue_number_goes_here> 🦕

### DIFF
--- a/google/api_core/path_template.py
+++ b/google/api_core/path_template.py
@@ -182,7 +182,7 @@ def get_field(request, field):
     Returns:
         The value of the field.
     """
-    parts = field.split('.')
+    parts = field.split(".")
     value = request
     for part in parts:
         if not isinstance(value, dict):
@@ -200,7 +200,7 @@ def delete_field(request, field):
         request (dict): A dictionary object.
         field (str): The key to the request in dot notation.
     """
-    parts = deque(field.split('.'))
+    parts = deque(field.split("."))
     while len(parts) > 1:
         if not isinstance(request, dict):
             return
@@ -210,7 +210,7 @@ def delete_field(request, field):
     if not isinstance(request, dict):
         return
     request.pop(part, None)
-    
+
 
 def validate(tmpl, path):
     """Validate a path against the path template.
@@ -235,6 +235,7 @@ def validate(tmpl, path):
     """
     pattern = _generate_pattern_for_template(tmpl) + "$"
     return True if re.match(pattern, path) is not None else False
+
 
 def transcode(http_options, **request_kwargs):
     """Transcodes a grpc request pattern into a proper HTTP request following the rules outlined here,
@@ -261,37 +262,39 @@ def transcode(http_options, **request_kwargs):
     """
     for http_option in http_options:
         request = {}
-        
+
         # Assign path
-        uri_template = http_option['uri']
-        path_fields = [match.group('name') for match in _VARIABLE_RE.finditer(uri_template)]
+        uri_template = http_option["uri"]
+        path_fields = [
+            match.group("name") for match in _VARIABLE_RE.finditer(uri_template)
+        ]
         path_args = {field: get_field(request_kwargs, field) for field in path_fields}
-        request['uri'] = expand(uri_template, **path_args)
+        request["uri"] = expand(uri_template, **path_args)
 
         # Remove fields used in uri path from request
         leftovers = copy.deepcopy(request_kwargs)
         for path_field in path_fields:
             delete_field(leftovers, path_field)
 
-        if not validate(uri_template, request['uri']) or not all(path_args.values()):
+        if not validate(uri_template, request["uri"]) or not all(path_args.values()):
             continue
 
         # Assign body and query params
-        body = http_option.get('body')
+        body = http_option.get("body")
 
         if body:
-            if body == '*':
-                request['body'] = leftovers
-                request['query_params'] = {}
+            if body == "*":
+                request["body"] = leftovers
+                request["query_params"] = {}
             else:
                 try:
-                    request['body'] = leftovers.pop(body)
+                    request["body"] = leftovers.pop(body)
                 except KeyError:
                     continue
-                request['query_params'] = leftovers
+                request["query_params"] = leftovers
         else:
-            request['query_params'] = leftovers
-        request['method'] = http_option['method']
+            request["query_params"] = leftovers
+        request["method"] = http_option["method"]
         return request
 
     raise ValueError("Request obj does not match any template")
--- a/google/api_core/path_template.py
+++ b/google/api_core/path_template.py
@@ -64,7 +64,7 @@ def _expand_variable_match(positional_vars, named_vars, match):
     """Expand a matched variable with its value.
 
     Args:
-        positional_vars (list): A list of positonal variables. This list will
+        positional_vars (list): A list of positional variables. This list will
             be modified.
         named_vars (dict): A dictionary of named variables.
         match (re.Match): A regular expression match.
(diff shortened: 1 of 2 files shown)


tseaver (commented):
@yihjenku Thank you very much for the patch!  I've tagged the Owlbot to run here, which should fix any linting issues (`black` wants spaces between key and value in dict comprehensions, for instance).  The testcases look well-thought out.

tests/unit/test_path_template.py
```diff
@@ -113,3 +113,161 @@ def test__replace_variable_with_pattern():
...
+            ],
+            {"field": "a/parent", "foo": "bar"},
+            ["get", "/v1/a/parent/b/first/last", {}, {"foo": "bar"}],
+        ],
+    ],
+)
+def test_transcode(http_options, request_kwargs, expected_result):
+    http_options, expected_result = helper_test_transcode(http_options, expected_result)
+    result = path_template.transcode(http_options, **request_kwargs)
+    assert result == expected_result
```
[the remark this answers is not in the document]
yihjenku: That's a good suggestion! I have split up the test into four use cases: base cases, wildcards, requests with an http body, and additional bindings. I think that should suffice!

google/api_core/path_template.py
```diff
@@ -193,3 +193,58 @@ def validate(tmpl, path):
...
+        Args:
+            http_options (list(dict)): A list of dicts which consist of these keys,
+                'method'    (str): The http method
+                'uri'       (str): The path template
+                'body'      (str): The body field name (optional)
+                (This is a simplified representation of the proto option `google.api.http`)
+            request_kwargs (dict) : A dict representing the request object
+
+        Returns:
+            dict: The transcoded request with these keys,
```
kbandes, line 210: I think it would be useful to include, in the return, the actual rule that matched, so somewhere below (say before line 248), you would have:
request['matched_rule'] = http_option

This would help the calling template figure out how to represent the returned body.  The body can't be used as-is because the names are snake case, and the actual json body needs to camel case, so the caller needs to manipulate it.

google/api_core/path_template.py
```diff
@@ -193,3 +193,58 @@ def validate(tmpl, path):
...
+        Raises:
+            ValueError: If the request does not match the given template.
+    """
+    request = {}
+    for http_option in http_options:
+
+        # Assign path
+        uri_template = http_option['uri']
+        path_fields = [match.group('name') for match in _VARIABLE_RE.finditer(uri_template)]
+        path_args = {field: request_kwargs.get(field) for field in path_fields}
```
kbandes, line 225: I think this will be a problem with path-type field names, like 'resource.type'.  If the request has a field 'resource' which is a message, containing a field 'type', the spec allows '{resource.type=*}'.  In this case, you need something like request_kwargs.get('request').get('type'), except more dynamic, something like:
val = request_kwargs
for name_seg = field.split('.'):
  val = val.get(name_seg)
but also look out for Nones, of course,  Maybe put it in a function.
There should also be test cases for this.

google/api_core/path_template.py
```diff
@@ -193,3 +193,58 @@ def validate(tmpl, path):
...
+        leftovers = {k: v for k,v in request_kwargs.items() if k not in path_args}
```
kbandes, line 226: This logic also doesn't handle subfields correctly.  If a template references `field.subfield`, e.g. `{field.subfield=*}`, the top level field is still in leftovers but the subfield should be removed from the body.

google/api_core/path_template.py
kbandes: To clarify:  consider this python snippet:
```
>>> x = {'a':{'b':'abc'},'c':'def'}
>>> def show_args(**kwargs):
...   print(kwargs)
... 
>>> show_args(**x)
{'a': {'b': 'abc'}, 'c': 'def'}
```
Note that you don't get anything like `a.b`, you get a nested dict (which makes sense).  So you have to navigate this in code.

google/api_core/path_template.py
```diff
@@ -193,3 +226,61 @@ def validate(tmpl, path):
...
+            dict: The transcoded request with these keys,
+                'method'        (str)   : The http method
+                'uri'           (str)   : The expanded uri
+                'body'          (dict)  : A dict representing the body (optional)
+                'query_params'  (dict)  : A dict mapping query parameter variables and values
+
+        Raises:
+            ValueError: If the request does not match the given template.
+    """
+    request = {}
```
jameslynnwu, line 252: let's move this into the http_options for loop to create a new request each time we consider a new http_option

google/api_core/path_template.py
```diff
@@ -193,3 +193,58 @@ def validate(tmpl, path):
...
+            ValueError: If the request does not match the given template.
+    """
+    answer = {}
+    for http_option in http_options:
+
+        # Assign path
+        uri_template = http_option['uri']
+        path_fields = [match.group('name') for match in _VARIABLE_RE.finditer(uri_template)]
+        path_args = {field:request_kwargs.get(field, None) for field in path_fields}
+        leftovers = {k:v for k,v in request_kwargs.items() if k not in path_args}
```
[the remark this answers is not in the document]
yihjenku: Addressed!

google/api_core/path_template.py
yihjenku: All set - I added the functions `get_field` and `delete_field` to address this problem. I also added unit tests.

google/api_core/path_template.py
yihjenku: Done.

google/api_core/path_template.py
yihjenku: Addressed.

google/api_core/path_template.py
yihjenku: As per offline discussion, not including this change in this pull request.

google/api_core/path_template.py
```diff
@@ -193,3 +226,62 @@ def validate(tmpl, path):
     """
     pattern = _generate_pattern_for_template(tmpl) + "$"
     return True if re.match(pattern, path) is not None else False
+
+def transcode(http_options, **request_kwargs):
+    """Transcodes a grpc request pattern into a proper HTTP request following the rules outlined here,
+       https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L44-L312
+
+        Args:
+            http_options (list(dict)): A list of dicts which consist of these keys,
```
alexander-fenster, line 235: Just to clarify: someone else will put the contents of `additional_binding` to this list, right?

alexander-fenster (approved):

google/api_core/path_template.py
```diff
@@ -193,3 +226,62 @@ def validate(tmpl, path):
...
+            dict: The transcoded request with these keys,
+                'method'        (str)   : The http method
+                'uri'           (str)   : The expanded uri
+                'body'          (dict)  : A dict representing the body (optional)
+                'query_params'  (dict)  : A dict mapping query parameter variables and values
+
+        Raises:
+            ValueError: If the request does not match the given template.
+    """
+    for http_option in http_options:
```
vam-google, line 252: I assume this function will be called for each request. It is pretty heavy one, and may potentially not satisfy  performance requirements imposed by Ads. But let's not do premature optimization here, so it is Ok.

vam-google (approved):
LGTM, but please keep in mind that we may have to circle back to optimize the implementation if it does not meet Ads performance requirements.

google/api_core/path_template.py
```diff
@@ -193,3 +226,62 @@ def validate(tmpl, path):
     """
     pattern = _generate_pattern_for_template(tmpl) + "$"
     return True if re.match(pattern, path) is not None else False
+
+def transcode(http_options, **request_kwargs):
+    """Transcodes a grpc request pattern into a proper HTTP request following the rules outlined here,
+       https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L44-L312
```
busunkim96, line 232: Would it be helpful at all to use the compiled `http_pb2.py` representation in the common protos package? https://github.com/googleapis/python-api-common-protos/tree/main/google/api

google/api_core/path_template.py
yihjenku: Yes correct.

google/api_core/path_template.py
yihjenku: I think it's ok for us to what's in `http.proto` for now, and we can look to use the compiled representation in `http_pb2.py` in the future.

google/api_core/path_template.py
```diff
@@ -170,6 +172,37 @@ def _generate_pattern_for_template(tmpl):
     return _VARIABLE_RE.sub(_replace_variable_with_pattern, tmpl)
 
 
+def get_field(request, field):
```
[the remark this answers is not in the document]
yihjenku: Done

google/api_core/path_template.py
```diff
@@ -170,6 +172,37 @@ def _generate_pattern_for_template(tmpl):
...
+        value = request
+        for part in parts:
+            if not isinstance(value, dict):
+                return
+            value = value[part]
+        if isinstance(value, dict):
+            return
+        return value
+    except KeyError:
+        return
```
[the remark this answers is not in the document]
yihjenku: Good tip, thanks! All set

google/api_core/path_template.py
```diff
@@ -170,6 +172,37 @@ def _generate_pattern_for_template(tmpl):
...
+            if not isinstance(request, dict):
+                return
+            part = parts.popleft()
+            request = request[part]
+        part = parts.popleft()
+        if not isinstance(request, dict):
+            return
+        del request[part]
+    except KeyError:
+        return
```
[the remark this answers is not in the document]
yihjenku: Done

busunkim96 (approved):

PR #272, opened 2021-09-02
chore: removing owlbot directives for conversion to main

nicain:
chore: removing owlbot directives for conversion to main


dandhlee (approved):
I can see that owlbot ran with no changes. LGTM 🦉

PR #275, opened 2021-09-16
feat: Add helper function to format query_params for rest transport.

kbandes:
This is a function needed to process the return values from the transcode function.

### DIFF
--- a/google/api_core/rest_helpers.py
+++ b/google/api_core/rest_helpers.py
@@ -1,0 +1,76 @@
+# Copyright 2017 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helpers for rest transports."""
+
+import itertools
+
+
+def flatten_query_params(obj, key_path=[]):
+    """Flatten a nested dict into a list of (name,value) tuples.
+
+    The result is suitable for setting query params on an http request.
+
+    .. code-block:: python
+
+        >>> obj = {'a':
+        ...         {'b':
+        ...           {'c': ['x', 'y', 'z']} },
+        ...      'd': 'uvw', }
+        >>> flatten_query_params(obj)
+        [('a.b.c', 'x'), ('a.b.c', 'y'), ('a.b.c', 'z'), ('d', 'uvw')]
+
+    Args:
+      obj: a nested dictionary (from json)
+      key_path: a list of name segments, representing levels above this obj.
+
+    Returns: a list of tuples, with each tuple having a (possibly) multi-part name
+      and a scalar value.
+    """
+
+    if obj is None:
+        return []
+    if isinstance(obj, dict):
+        return _flatten_dict(obj, key_path=key_path)
+    if isinstance(obj, list):
+        return _flatten_list(obj, key_path=key_path)
+    return _flatten_value(obj, key_path=key_path)
+
+
+def _is_value(obj):
+    if obj is None:
+        return False
+    return not (isinstance(obj, list) or isinstance(obj, dict))
+
+
+def _flatten_value(obj, key_path=[]):
+    if not key_path:
+        # There must be a key.
+        return []
+    return [('.'.join(key_path), obj)]
+
+
+def _flatten_dict(obj, key_path=[]):
+    return list(
+        itertools.chain(*(flatten_query_params(v, key_path=key_path + [k])
+                        for k, v in obj.items())))
+
+
+def _flatten_list(l, key_path=[]):
+    # Only lists of scalar values are supported.
+    # The name (key_path) is repeated for each value.
+    return list(
+        itertools.chain(*(_flatten_value(elem, key_path=key_path)
+                          for elem in l
+                          if _is_value(elem))))
--- a/google/api_core/rest_helpers.py
+++ b/google/api_core/rest_helpers.py
@@ -75,9 +75,6 @@ def _is_primitive_value(obj):
 
 
 def _flatten_value(obj, key_path):
-    if not key_path:
-        # There must be a key.
-        return []
     return [('.'.join(key_path), obj)]
 
 
--- a/google/api_core/rest_helpers.py
+++ b/google/api_core/rest_helpers.py
@@ -75,14 +75,11 @@ def _is_primitive_value(obj):
 
 
 def _flatten_value(obj, key_path):
-    return [('.'.join(key_path), obj)]
+    return [(".".join(key_path), obj)]
 
 
 def _flatten_dict(obj, key_path):
-    items = (
-        _flatten(value, key_path=key_path + [key])
-        for key, value in obj.items()
-    )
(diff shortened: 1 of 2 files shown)


google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+        return _flatten_dict(obj, key_path=key_path)
+    if isinstance(obj, list):
+        return _flatten_list(obj, key_path=key_path)
+    return _flatten_value(obj, key_path=key_path)
+
+
+def _is_value(obj):
+    if obj is None:
+        return False
+    return not (isinstance(obj, list) or isinstance(obj, dict))
```
tseaver, line 54: This can be simplified to:
```suggestion
    return obj is not None and not isinstance(obj, (list, dict))
```

google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def _flatten_value(obj, key_path=[]):
```
tseaver, line 57: These are private helpers:  we know that the callers always pass appropriate `key_path` values, so they don't need to be optional:
```suggestion
def _flatten_value(obj, key_path):
```

google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    if not key_path:
+        # There must be a key.
+        return []
```
tseaver, line 60: This never happens:  the caller always passes a real path:
```suggestion
```

google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    return [('.'.join(key_path), obj)]
+
+
+def _flatten_dict(obj, key_path=[]):
```
tseaver, line 64: LIkewise:
```suggestion
def _flatten_dict(obj, key_path):
```

google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    return list(
+        itertools.chain(*(flatten_query_params(v, key_path=key_path + [k])
+                        for k, v in obj.items())))
+
+
+def _flatten_list(l, key_path=[]):
```
tseaver, line 70: And here, both avoid the one-character parameter name and the optional value:
```suggestion
def _flatten_list(elems, key_path):
```

google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    # Only lists of scalar values are supported.
+    # The name (key_path) is repeated for each value.
+    return list(
+        itertools.chain(*(_flatten_value(elem, key_path=key_path)
+                          for elem in l
+                          if _is_value(elem))))
```
tseaver, line 76: Silently dropping non-leaf elements seems surprising:  I would expect an exception if I passed through something which couldn't be flattened properly.  Also, whichever semantic is appropriate should be explicitly documented in the docstring for `flatten_query_params`.

google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+def _flatten_value(obj, key_path=[]):
+    if not key_path:
+        # There must be a key.
+        return []
+    return [('.'.join(key_path), obj)]
+
+
+def _flatten_dict(obj, key_path=[]):
+    return list(
+        itertools.chain(*(flatten_query_params(v, key_path=key_path + [k])
```
kbandes, line 66: I could also do this with a loop, extending a list with each return from flatten_query_params.  I'm open to suggestion here.

google/api_core/rest_helpers.py
tseaver: Note as well that we need to exercise each non-value case in the unit tests.

google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+                        for k, v in obj.items())))
```
tseaver, line 67: Please use `key` and `value` rather than `k`, `v`:
```suggestion
        itertools.chain(*(flatten_query_params(value, key_path=key_path + [key])
                        for key, value in obj.items())))
```

google/api_core/rest_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def _flatten_list(l, key_path=[]):
+    # Only lists of scalar values are supported.
+    # The name (key_path) is repeated for each value.
+    return list(
+        itertools.chain(*(_flatten_value(elem, key_path=key_path)
+                          for elem in l
```
tseaver, line 75:
```suggestion
                          for elem in elems
```

tseaver (changes requested):
Please run `nox -e blacken lint` and push the changes.

google/api_core/rest_helpers.py
tseaver: I found the `itertools.chain` bit hard to understand.  I would've reached for `functools.reduce`, myself, e.g.
```python
    items = (
        flatten_query_params(value, key_path=key_path + [key])
        for key, value in obj.items()
    )
    return functools.reduce(operator.concat, items, [])
```

google/api_core/rest_helpers.py
tseaver: As above:
```python
    items = (
        _flatten_value(elem, key_path=key_path) for elem in elems if _is_value(elem)
    )
    return functools.reduce(operator.concat, items, [])
```

google/api_core/rest_helpers.py
kbandes: _flatten_value can be called from flatten_query_params, which should be called with a dict, but in theory could be called as `flatten_query_params("abc")`.

google/api_core/rest_helpers.py
kbandes: done

google/api_core/rest_helpers.py
kbandes: done

google/api_core/rest_helpers.py
kbandes: done

google/api_core/rest_helpers.py
kbandes: done

google/api_core/rest_helpers.py
tseaver: What would the caller's intent be there?  This is another case where having the function return an empty list silently is surprising:  `flatten_query_params("abc")` should probably raise an exception.

kbandes: That's fair.  I factored out a top-level function to check the top-level type, from the helper function to check inner types.  The top level raises an exception if the input is not a dict or None.

google/api_core/rest_helpers.py
kbandes: Done - it raises an exception (ValueError) for repeated, non-primitive fields.

I distinguish calling the top-level function with a non-dict (TypeError) from this case.  The containing type is correct, but the inner type is not (in this context).  So I consider this a bad value for a good type (dict at the top level).

google/api_core/rest_helpers.py
kbandes: btw, strictly speaking this could use operator.add, since items are operated upon pairwise.

google/api_core/rest_helpers.py
kbandes: done

kbandes:
Honestly, it's annoying that black and autopep8 have different ideas about what is proper formatting, and different repos prefer one or the other.  But oh well.

google/api_core/rest_helpers.py
tseaver: Given that it is unlikely anyone would be catching either of those exceptions (they represent programming errors, not operational ones), that choice makes sense to me.

tseaver (approved):

PR #286, opened 2021-10-08
feat: Add error_details property to GoogleAPICallError based on google.rpc.status.details.

### LINKED ISSUE #99: Use package grpcio-status to parse rich errors in trailing metadata.
Package: https://pypi.org/project/grpcio-status/

Example: https://github.com/grpc/grpc/blob/master/examples/python/errors/client.py#L37

> We are from IAM Policy Analyzer team and we've recently found that the python client library does not seems to surface the `details` in the Error Status properly. We've tried using gcloud and it works as expected. The comparison is as below:

```
$ gcloud beta asset analyze-iam-policy --project=cai-playground
ERROR: (gcloud.beta.asset.analyze-iam-policy) INVALID_ARGUMENT: Some specified value(s) are invalid.
- '@type': type.googleapis.com/google.rpc.BadRequest
  fieldViolations:
  - description: At least one of resource selector, identity selector or access selector
      needs to  be specified.
    field: analysis_query
```

```py
$ gcloud auth application-default login
$ python3
 from google.cloud import asset_v1p4beta1
 from google.cloud.asset_v1p4beta1 import AnalyzeIamPolicyRequest, IamPolicyAnalysisQuery
 parent="projects/cai-playground"
 client = asset_v1p4beta1.AssetServiceClient()
 response = client.analyze_iam_policy(request=AnalyzeIamPolicyRequest(analysis_query=IamPolicyAnalysisQuery(parent=parent)))

Traceback (most recent call last):
  File "/usr/local/google/home/aaronlichen/.local/lib/python3.8/site-packages/google/api_core/grpc_helpers.py", line 57, in error_remapped_callable
    return callable_(*args, **kwargs)
  File "/usr/local/google/home/aaronlichen/.local/lib/python3.8/site-packages/grpc/_channel.py", line 826, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "/usr/local/google/home/aaronlichen/.local/lib/python3.8/site-packages/grpc/_channel.py", line 729, in _end_unary_response_blocking
    raise _InactiveRpcError(state)
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.INVALID_ARGUMENT
        details = "Some specified value(s) are invalid."
        debug_error_string = "{"created":"@1602024200.721639270","description":"Error received from peer ipv4:74.125.142.95:443","file":"src/core/lib/surface/call.cc","file_line":1061,"grpc_message":"Some specified value(s) are invalid.","grpc_status":3}"
```

atulep:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-api-core/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [?] Code coverage does not decrease (if any source code was changed)
- [?] Appropriate docs were updated (if necessary)

Fixes #99 🦕


google/api_core/exceptions.py
```diff
@@ -22,12 +22,15 @@
 from __future__ import unicode_literals
 
 import http.client
+from google.rpc import error_details_pb2
```
tseaver, line 25: Two things:

- `grpcio` is a conditional dependency (some REST-only libraries don't need or want it).
- We still support older versions of it that don't have the `details_pb2` module (down to `grpcio 1.33.2`).

This import needs to be moved inside its own `try:` (because of older versions) below, and a default value set in the `except ImportError:` block.

`google.rpc`

google/api_core/exceptions.py
```diff
@@ -22,12 +22,15 @@
...
 
 try:
     import grpc
+    from grpc_status import rpc_status
```
tseaver, line 29: Do we know for sure that `grpc_status` is compatible with the oldest version of `grpcio` we support?  If not, this import needs to be inside its own `try: .. except ImportError:` block, and we would also need to add it conditionally as another extra in `setup.py`.

google/api_core/exceptions.py
```diff
@@ -461,6 +480,41 @@ def _is_informative_grpc_error(rpc_exc):
...
+        error_details_pb2.PreconditionFailure,
+        error_details_pb2.QuotaFailure,
+        error_details_pb2.ErrorInfo,
+        error_details_pb2.RetryInfo,
+        error_details_pb2.ResourceInfo,
+        error_details_pb2.RequestInfo,
+        error_details_pb2.DebugInfo,
+        error_details_pb2.Help,
+        error_details_pb2.LocalizedMessage,
+    ]
```
tseaver, line 502: Can we compute this list coditionally as `[]` if `details_pb2` is not importable?

setup.py
```diff
@@ -36,7 +36,7 @@
     "setuptools >= 40.3.0",
 ]
 extras = {
-    "grpc": "grpcio >= 1.33.2, < 2.0dev",
+    "grpc": ["grpcio >= 1.33.2, < 2.0dev", "grpcio-status >= 1.0.0, < 2.0dev"],
```
tseaver, line 39: May need to be its own extra (see my comment above).

tests/unit/test_exceptions.py
```diff
@@ -14,35 +14,37 @@
...
 import http.client
 import json
+from google.rpc import status_pb2
+from google.rpc.status_pb2 import Status
 
 import grpc
 import mock
 import requests
-
+from google.rpc import error_details_pb2
```
tseaver, line 23: This import needs to be conditional:
```suggestion
try:
    from google.rpc import error_details_pb2
except ImportError:
    details_pb2 = None
```

tests/unit/test_exceptions.py
```diff
@@ -14,35 +14,37 @@
...
+from google.rpc import code_pb2
+from google.protobuf import any_pb2, json_format
+from grpc_status import rpc_status
```
tseaver, line 26: Likewise, these imports should be conditional.

google/api_core/exceptions.py
atulep: +1 - `error_details_pb2` comes from [googleapis-common-protos](https://github.com/googleapis/python-api-common-protos/tree/main/google/rpc).

tests/unit/test_exceptions.py
atulep: The only conditional import should be `grpc_status` right? Assuming `error_details_pb2` and `any_pb2` come from `googleapis-common-protos`; `google.protobuf` is a common dependency.

google/api_core/exceptions.py
```diff
@@ -116,15 +119,16 @@ class GoogleAPICallError(GoogleAPIError, metaclass=_GoogleAPICallErrorMeta):
     This may be ``None`` if the exception does not match up to a gRPC error.
     """
 
-    def __init__(self, message, errors=(), response=None):
+    def __init__(self, message, errors=(), details=(), response=None):
```
tseaver, line 122: We need a corresponding update to the class docstring above.

google/api_core/exceptions.py
```diff
@@ -116,15 +119,16 @@ class GoogleAPICallError(GoogleAPIError, metaclass=_GoogleAPICallErrorMeta):
...
         super(GoogleAPICallError, self).__init__(message)
         self.message = message
         """str: The exception message."""
         self._errors = errors
+        self._details = details
         self._response = response
 
     def __str__(self):
-        return "{} {}".format(self.code, self.message)
+        return "{} {} {}".format(self.code, self.message, self.error_details)
```
tseaver, line 131: Having the empty list show up is distracting and not useful.  Better would be:
```suggestion
        details = self.error_details
        if details:
            return "{} {} {}".format(self.code, self.message, details)
        else:
            return "{} {}".format(self.code, self.message)
```

tests/unit/test_exceptions.py
tseaver: Strictly, the `grpc` import should also be conditional.  Also, all the non-stdlib and non-`google.api_core` imports should be grouped together and sorted.

tests/unit/test_exceptions.py
```diff
@@ -14,35 +14,37 @@
...
 from google.api_core import exceptions
-
+from google.protobuf import text_format
 
 def test_create_google_cloud_error():
     exception = exceptions.GoogleAPICallError("Testing")
     exception.code = 600
-    assert str(exception) == "600 Testing"
+    assert str(exception) == "600 Testing []"
```
tseaver, line 33: As above, the empty list is unhelpful.  We also need a test which exercises the `__str__` method when there *are* error details present.

tseaver (changes requested):

google/api_core/exceptions.py
```diff
@@ -461,6 +480,41 @@ def _is_informative_grpc_error(rpc_exc):
     return hasattr(rpc_exc, "code") and hasattr(rpc_exc, "details")
 
 
+def _parse_grpc_error_details(rpc_exc):
+    if not rpc_status:
+        return []
```
tseaver, line 485: No coverage for this case.

google/api_core/exceptions.py
atulep: Done

tests/unit/test_exceptions.py
atulep: Done. I added a check in a different test later in the code.

testing/constraints-3.6.txt
```diff
@@ -14,3 +14,4 @@ packaging==14.3
 grpcio==1.33.2
 grpcio-gcp==0.2.2
 grpcio-gcp==0.2.2
+grpcio-status==1.33.2
```
busunkim96, line 17:
```suggestion
grpcio-status==1.0.0
```

(to match the lower bound in setup.py)

tseaver: Isn't the `grpcio-status` version supposed to be identical to the one for `grpcio`?

google/api_core/exceptions.py
```diff
@@ -135,6 +143,19 @@ def errors(self):
         """
         return list(self._errors)
 
+    @property
+    def error_details(self):
```
summer-ji-eng, line 147: Is `details` property has been occupied?  If not, is that better to use `details` as name or `status_details` relate to `google.rpc.Status.details`. 
Notice that we need to keep consistency between grpc and REST.
IMO, I prefer `status_details`. WDYT @busunkim96 @tseaver @atulep

summer-ji-eng (commented):
First of all, thank you @atulep a lot working on python error improvement. It provide better debugging experience to the developer who use python GAPIC client libraries. 🚀 

I leave the python language specific review with python expertise. For the logic of unpacking error details part it looks good to me. Couple things to double confirm:
1. This implementation include both grpc and REST. I see `_parse_grpc_error_details` and `from_http_response` to get status.details. (That's great!)
2. This implementation doesn't promote the ErrorInfo fields to the top level of `exception`, are you planning to do that in separated PR? Please let me know if you have any question on this.
3. Last but not least, could you add an integration test against `Showcase`. @busunkim96 @software-dov should have ideas where to add IT.

tests/unit/test_exceptions.py
```diff
@@ -235,3 +235,89 @@ def test_from_grpc_error_non_call():
...
+    response = make_response(
+        json.dumps(
+            {"error": {"message": "\u2019 message", "errors": ["1", "2"]}}
+        ).encode("utf-8")
+    )
+    exception = exceptions.from_http_response(response)
+    assert exception.details == []
+
+
+def test_error_details_from_grpc_response():
```
tseaver, line 288: Now that #289 has landed on `main`, we need to skip this test when `grpc` cannot be imported.
```suggestion
@pytest.mark.skipif(grpc is None, reason="gRPC not importable")
def test_error_details_from_grpc_response():
```

tests/unit/test_exceptions.py
```diff
@@ -235,3 +235,89 @@ def test_from_grpc_error_non_call():
...
+    with mock.patch("grpc_status.rpc_status.from_call") as m:
+        m.return_value = status
+        exception = exceptions.from_grpc_error(error)
+
+    bad_request_detail = error_details_pb2.BadRequest()
+    status_detail.Unpack(bad_request_detail)
+    assert exception.details == [bad_request_detail]
+
+
+def test_error_details_from_grpc_response_unknown_error():
```
tseaver, line 309: Likewise here:
```suggestion
@pytest.mark.skipif(grpc is None, reason="gRPC not importable")
def test_error_details_from_grpc_response_unknown_error():
```

tseaver (changes requested):

tseaver (approved):

PR #290, opened 2021-10-13
fix: add mypy checking + 'py.typed' file

tseaver:


noxfile.py
```diff
@@ -155,6 +156,14 @@ def pytype(session):
     session.run("pytype")
 
 
+@nox.session(python="3.6")
```
tswast, line 159: Why not DEFAULT_PYTHON_VERSION?

tseaver: I'm matching what is in the `mypy.ini` file I lifted from @crwilcox.

crwilcox: that specifies to mypy what minimum version of python the types should be compatible with. If I understand correctly we should be able to run against a later python version

tswast (approved):

PR #292, opened 2021-10-14
fix: make 'gapic_v1.method.DEFAULT' a typed object

tseaver:
FBO checkers which need to verify default values for 'retry' / 'timeout'.

### DIFF
--- a/google/api_core/gapic_v1/method.py
+++ b/google/api_core/gapic_v1/method.py
@@ -18,6 +18,7 @@ This is used by gapic clients to provide common error mapping, retry, timeout,
 pagination, and long-running operations to gRPC methods.
 """
 
+import enum
 import functools
 
 from google.api_core import grpc_helpers
@@ -25,7 +26,18 @@ from google.api_core import timeout
 from google.api_core.gapic_v1 import client_info
 
 USE_DEFAULT_METADATA = object()
-DEFAULT = object()
+
+
+class _MethodDefault(enum.Enum):
+    # Uses enum so that pytype/mypy knows that this is the only possible value.
+    # https://stackoverflow.com/a/60605919/101923
+    #
+    # Literal[_DEFAULT_VALUE] is an alternative, but only added in Python 3.8.
+    # https://docs.python.org/3/library/typing.html#typing.Literal
+    _DEFAULT_VALUE = object()
+
+
+DEFAULT = _MethodDefault._DEFAULT_VALUE
 """Sentinel value indicating that a retry or timeout argument was unspecified,
 so the default should be used."""
 


google/api_core/gapic_v1/method.py
```diff
@@ -18,14 +18,26 @@
...
 
 from google.api_core import grpc_helpers
 from google.api_core import timeout
 from google.api_core.gapic_v1 import client_info
 
 USE_DEFAULT_METADATA = object()
-DEFAULT = object()
+
+
+class _MethodDefault(enum.Enum):
```
tswast, line 31: Since this needs to be used in the generated code, should we make this public?

tseaver: I figured that the generator is in the "friends zone" for api-core. :)

crwilcox: I think the concern is that this type will be exposed as a type to users of the API?

tseaver: It seems like an "implementation detail" to me, uninteresting to anyone.  It will be in generated code, and present only in the definition of the `OptionalRetry` type -- see https://github.com/googleapis/gapic-generator-python/pull/1032#pullrequestreview-780331392.  I am in fact considering a change to that PR which removes even the classname.  Instead of:

```python
try:
    OptionalRetry = Union[retries.Retry, gapic_v1.method._MethodDefault]
except AttributeError:
    OptionalRetry = Union[retries.Retry, object]
```

I'm thinking this would be simpler / clearer:
```python
OptionalRetry = Union[retries.Retry, type(gapic_v1.method.DEFAULT)]
```

tswast: > I think the concern is that this type will be exposed as a type to users of the API?

Yes, that is a concern.

On second thought though, I think most users usages of the `_MethodDefault` type will be by omitting the `retry` argument, so they probably won't need to access this directly. Especially not if `OptionalRetry` is exposed in the generated client.

tswast (approved):

tswast:
Adding Do Not Merge so Chris and/or Bu Sun can approve.

tseaver:
@tswast Does the merge of https://github.com/googleapis/gapic-generator-python/commit/d85dfadc180e5f218ad582a306c1c441a6c668db imply that we need to merge here, too?

tswast (approved):

tswast:
> Does the merge of googleapis/gapic-generator-python@d85dfad imply that we need to merge here, too?

Yes, let's do it. We'll need to cut a release too. Thankfully since there's a fallback in the generated client, we should be fine if releases are out-of-sync.

PR #311, opened 2021-11-19
feat: add operations rest client to support long-running operations.

kbandes:
This PR adds a rest client for the Operations service under each API.

This client is a modified version of the auto-generated gapic client for google.longrunning.Operation.  The modifications make the operations rest client method signatures compatible with the grpc client.

### DIFF
--- a/google/api_core/operations_v1/__init__.py
+++ b/google/api_core/operations_v1/__init__.py
@@ -16,5 +16,12 @@
 
 from google.api_core.operations_v1.operations_async_client import OperationsAsyncClient
 from google.api_core.operations_v1.operations_client import OperationsClient
+from google.api_core.operations_v1.transports.rest import OperationsRestTransport
+from google.api_core.operations_v1.operations_rest_client import OperationsRestClient
 
-__all__ = ["OperationsAsyncClient", "OperationsClient"]
+__all__ = [
+    "OperationsAsyncClient",
+    "OperationsClient",
+    "OperationsRestClient",
+    "OperationsRestTransport"
+]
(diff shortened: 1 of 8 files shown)


software-dov (commented):
This is massively, massively outside the scope of the review, but we could just gapic-ify the entire operations client and give it both grpc and rest transports. Then the 'normal' gapic would be changed to instantiate an operations client with the same transport type as itself.

I'm not sure that's possible without making a breaking change, but we could try to mitigate it a bit, or at least make any eventual transition easier by renaming OperationsRest* to OperationsAbstract* or similar. This may not be worth it, though.

kbandes:
Right, OperationsRestClient is not really "rest" specific, and transports/base.py isn't either.

There are several obstacles to using a completely auto-generated gapic client for Operations, though:
1.  I changed ALL the api calls to have the same signature as the existing grpc operations client.  This means individual field parameters instead of request objects, and return values of the pb2-style wrapper for `google.longrunning.Operations`, rather than the proto+ wrapper.  Using the default, auto-generated method signatures would be breaking code for any existing applications that call the Operations client directly.
2. For rest in particular, I had to hand-punch a hole for alternate http specifications to drop into the Operations rest transport.  Each API has its own versions of these things for the operations methods.  That's why I needed to add the service yaml stuff in the generator.

software-dov:
> alternate http specifications to drop into the Operations rest transport

This right here seems to be a showstopper. Okay, I'm convinced that this would be a seriously non-trivial thing to do. Let's leave it as a future or never problem, then.

kbandes:
I'm happy to rename the client class though.

kbandes:
Note that I will need to update #1094 in googleapis/gapic-generator-python for this change.

software-dov (approved):

PR #307, opened 2021-11-08
feat: add support for 'error_info'

atulep:
Adds `error_info` field to `GoogleAPICallError`, as requested in googleapis/python-api-core#286.

PR #315, opened 2021-11-29
feat: add support for 'error_info'

atulep:
Adds `error_info` field to `GoogleAPICallError`, as requested in #286.

This is a duplicate of #307, except now CLA won't prevent from merging.

### DIFF
--- a/google/api_core/exceptions.py
+++ b/google/api_core/exceptions.py
@@ -472,8 +472,13 @@ def from_http_response(response):
     errors = payload.get("error", {}).get("errors", ())
     # In JSON, details are already formatted in developer-friendly way.
     details = payload.get("error", {}).get("details", ())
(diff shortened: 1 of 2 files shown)


tseaver (approved):

google/api_core/exceptions.py
```diff
@@ -433,13 +472,21 @@ def from_http_response(response):
     errors = payload.get("error", {}).get("errors", ())
     # In JSON, details are already formatted in developer-friendly way.
     details = payload.get("error", {}).get("details", ())
+    error_info_type = "type.googleapis.com/google.rpc.ErrorInfo"
+    error_info = [d for d in details if d.get("@type", "") == error_info_type]
```
tseaver, line 476: The `pytype` failure here is puzzling.  Perhaps we need to add a type assertion for either `payload` or `details`?

atulep: I refactored this code.

PR #317, opened 2021-12-10
feat: iterator for processing JSON responses in REST streaming.

atulep:
Implements an iterator that processes an array of JSON responses coming from REST server-side streaming API.

### DIFF
--- a/google/api_core/rest_streaming.py
+++ b/google/api_core/rest_streaming.py
@@ -14,8 +14,10 @@
 
 """Helpers for server-side streaming in REST."""
 
+from collections import deque
 import json
 import string
+from typing import Deque
 
 import requests
 
@@ -35,7 +37,7 @@ class ResponseIterator:
         # Inner iterator over HTTP response's content.
         self._response_itr = self._response.iter_content(decode_unicode=True)
         # Contains a list of JSON responses ready to be sent to user.
-        self._ready_objs = []
+        self._ready_objs: Deque[str] = deque()
         # Current JSON response being built.
         self._obj = ""
         # Keeps track of the nesting level within a JSON object.
@@ -96,10 +98,10 @@ class ResponseIterator:
         return self._grab()
 
     def _grab(self):
-        obj = self._ready_objs[0]
-        self._ready_objs = self._ready_objs[1:]
         # Add extra quotes to make json.loads happy.
-        return self._response_message_cls.from_json(json.loads('"' + obj + '"'))
+        return self._response_message_cls.from_json(
+            json.loads('"' + self._ready_objs.popleft() + '"')
+        )
 
     def __iter__(self):
         return self
--- a/google/api_core/rest_streaming.py
+++ b/google/api_core/rest_streaming.py
@@ -77,9 +77,15 @@ class ResponseIterator:
                 if self._in_string:
                     self._obj += char
             elif char == "[":
-                self._level += 1
+                if self._level == 0:
+                    self._level += 1
+                else:
+                    self._obj += char
             elif char == "]":
-                self._level -= 1
+                if self._level == 1:
+                    self._level -= 1
+                else:
+                    self._obj += char
             else:
                 self._obj += char
 
--- a/google/api_core/rest_streaming.py
+++ b/google/api_core/rest_streaming.py
@@ -63,14 +63,16 @@ class ResponseIterator:
                 if self._level == 1:
                     # Level 1 corresponds to the outermost JSON object
                     # (i.e. the one we care about).
-                    self._obj = char
+                    self._obj = ""
                 if not self._in_string:
                     self._level += 1
+                self._obj += char
             elif char == "}":
+                self._obj += char
                 if not self._in_string:
                     self._level -= 1
                 if not self._in_string and self._level == 1:
-                    self._ready_objs.append(self._obj + char)
+                    self._ready_objs.append(self._obj)
             elif char == '"':
                 # Helps to deal with an escaped quotes inside of a string.
                 if not self._next_should_be_escaped:
@@ -114,6 +116,7 @@ class ResponseIterator:
 
     def _grab(self):
         # Add extra quotes to make json.loads happy.
+        print(self._ready_objs)
         return self._response_message_cls.from_json(self._ready_objs.popleft())
 
     def __iter__(self):
--- a/google/api_core/rest_streaming.py
+++ b/google/api_core/rest_streaming.py
@@ -15,7 +15,6 @@
 """Helpers for server-side streaming in REST."""
 
 from collections import deque
-import json
 import string
 from typing import Deque
 
@@ -45,6 +44,8 @@ class ResponseIterator:
         # Keeps track whether HTTP response is currently sending values
         # inside of a string value.
         self._in_string = False
+        # Whether an escape symbol "\" was encountered.
+        self._next_should_be_escaped = False
 
     def cancel(self):
         """Cancel existing streaming operation.
@@ -62,19 +63,19 @@ class ResponseIterator:
                 if self._level == 1:
                     # Level 1 corresponds to the outermost JSON object
                     # (i.e. the one we care about).
-                    self._obj = ""
+                    self._obj = char
                 if not self._in_string:
                     self._level += 1
-                self._obj += char
-            elif char == '"':
-                self._in_string = not self._in_string
-                self._obj += char
             elif char == "}":
-                self._obj += char
                 if not self._in_string:
                     self._level -= 1
                 if not self._in_string and self._level == 1:
-                    self._ready_objs.append(self._obj)
+                    self._ready_objs.append(self._obj + char)
+            elif char == '"':
+                # Helps to deal with an escaped quotes inside of a string.
+                if not self._next_should_be_escaped:
+                    self._in_string = not self._in_string
+                self._obj += char
             elif char in string.whitespace:
                 if self._in_string:
                     self._obj += char
@@ -91,17 +92,29 @@ class ResponseIterator:
             else:
                 self._obj += char
 
+            if char == "\\":
+                # Escaping the "\".
+                if self._next_should_be_escaped:
+                    self._next_should_be_escaped = False
+                else:
+                    self._next_should_be_escaped = True
+            else:
+                self._next_should_be_escaped = False
+
     def __next__(self):
         while not self._ready_objs:
-            chunk = next(self._response_itr)
-            self._process_chunk(chunk)
+            try:
+                chunk = next(self._response_itr)
+                self._process_chunk(chunk)
+            except StopIteration as e:
+                if self._level > 0:
+                    raise ValueError("Unfinished stream: %s" % self._obj)
+                raise e
         return self._grab()
 
     def _grab(self):
         # Add extra quotes to make json.loads happy.
-        return self._response_message_cls.from_json(
-            json.loads('"' + self._ready_objs.popleft() + '"')
-        )
+        return self._response_message_cls.from_json(self._ready_objs.popleft())
 
     def __iter__(self):
         return self
--- a/google/api_core/rest_streaming.py
+++ b/google/api_core/rest_streaming.py
@@ -116,7 +116,6 @@ class ResponseIterator:
 
     def _grab(self):
         # Add extra quotes to make json.loads happy.
-        print(self._ready_objs)
         return self._response_message_cls.from_json(self._ready_objs.popleft())
 
     def __iter__(self):
--- a/google/api_core/rest_streaming.py
+++ b/google/api_core/rest_streaming.py
@@ -1,0 +1,99 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


google/api_core/rest_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+                    self._ready_objs.append(self._obj)
+            elif char in string.whitespace:
+                if self._in_string:
+                    self._obj += char
+            elif char == "[":
+                self._level += 1
+            elif char == "]":
+                self._level -= 1
+            else:
+                self._obj += char
```
vam-google, line 84: Looks like it is the most common case, but it comes  last in the whole if-else huge chain, which means almost for every character in the stream we will have to go through every if-else block before reaching the final destination. It is not a very big deal these days, but still feels a bit inefficient. 

Optional: Python does not have proper switch stateements to deal with it, but we can do the a dict instead of if-else. I.e. `if char in whitespace else if char not in special_chars else self._obj += char`, where `special_chars = "{}\"[]"`.

google/api_core/rest_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+        for char in chunk:
+            if char == "{":
+                if self._level == 1:
+                    # Level 1 corresponds to the outermost JSON object
+                    # (i.e. the one we care about).
+                    self._obj = ""
+                if not self._in_string:
+                    self._level += 1
+                self._obj += char
+            elif char == '"':
```
vam-google, line 67: Do single qoutes need any special parsing?

tests/unit/test_rest_streaming.py
```diff
@@ -0,0 +1,160 @@
...
+
+from google.api_core import rest_streaming
+
+
+class Composer(proto.Message):
+    given_name = proto.Field(proto.STRING, number=1)
+    family_name = proto.Field(proto.STRING, number=2)
+
+
+class Song(proto.Message):
```
vam-google, line 32: Can we also add special fields like:
int64 (serialized as string, parsed as int)
[Int32Value](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto#L88)
Duration
Timestamp

Full list of special JSON mappings according to proto spec can be found here: https://developers.google.com/protocol-buffers/docs/proto3#json

google/api_core/rest_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+        # Keeps track whether HTTP response is currently sending values
+        # inside of a string value.
+        self._in_string = False
+
+    def cancel(self):
+        """Cancel existing streaming operation.
+        """
+        self._response.close()
+
+    def _process_chunk(self, chunk: str):
```
vam-google, line 52: Is it safe to assume that this code block is 1-1 translation from the TypeScript implementation?

vam-google (commented):
Overall looks good. Just a few questions/suggestions.

google/api_core/rest_streaming.py
atulep: No, as JSON standard disallows single quotes - https://stackoverflow.com/questions/36038454/parsing-string-as-json-with-single-quotes

google/api_core/rest_streaming.py
atulep: They're very close. I had to change some data structures (i.e. `self._ready_objs`) because TypeScript follows a specific Node framework that handles some of housekeeping.

google/api_core/rest_streaming.py
atulep: Is performance your main concern? I'm not sure if reordering of if-statements will make a significant difference to end user because network I/O is involved. If you insist, I can try reducing if-statements with dict as you suggested.

google/api_core/rest_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+        class expected to be returned from an API.
+    """
+
+    def __init__(self, response: requests.Response, response_message_cls):
+        self._response = response
+        self._response_message_cls = response_message_cls
+        # Inner iterator over HTTP response's content.
+        self._response_itr = self._response.iter_content(decode_unicode=True)
+        # Contains a list of JSON responses ready to be sent to user.
+        self._ready_objs = []
```
software-dov, line 38: Nit: consider using a `deque` since we're removing objects in insertion order.

tests/unit/test_rest_streaming.py
atulep: protoplus doesn't have Duration and Timestamp, but I've added other fields.

tests/unit/test_rest_streaming.py
```diff
@@ -0,0 +1,160 @@
...
+            self._i = 0
+            self._decode_unicode = decode_unicode
+            self._random_split = random_split
+            random.seed(seed)
+
+        def __next__(self):
+            if self._i == len(self._responses_bytes):
+                raise StopIteration
+            if self._random_split:
+                n = random.randint(1, len(self._responses_bytes[self._i :]))
```
software-dov, line 70: I think a randomized test is a good idea, but I also think it's important to keep track of the random seed in case we start seeing sporadic test failures. Maybe something like:

```python
import logging
import time

SEED = time.time_ns()
logging.info(f"Starting rest streaming tests with random seed: {SEED}")
random.seed(SEED)
```

tests/unit/test_rest_streaming.py
```diff
@@ -0,0 +1,160 @@
...
+        # json.dumps returns a string surrounded with quotes that need to be stripped
+        # in order to be an actual JSON.
+        json_responses = [
+            json.dumps(self._response_message_cls.to_json(r))[1:-1] for r in responses
+        ]
+        for x in json_responses:
+            ret_val += x
+            ret_val += ","
+        ret_val = ret_val[:-1]  # Remove trailing comma.
+        ret_val += "]"
```
software-dov, line 102: The python idiom to do exactly this is:
```python
ret_val = "[{}]".format(",".join(json_responses))
```

It strips the trailing comma for you.

tests/unit/test_rest_streaming.py
```diff
@@ -0,0 +1,160 @@
...
+        self._responses = responses
+        self._random_split = random_split
+        self._response_message_cls = response_cls
+
+    def _parse_responses(self, responses: List[proto.Message]) -> bytes:
+        ret_val = "["
+        # json.dumps returns a string surrounded with quotes that need to be stripped
+        # in order to be an actual JSON.
+        json_responses = [
+            json.dumps(self._response_message_cls.to_json(r))[1:-1] for r in responses
```
software-dov, line 96: What do you think about
```python
json.dumps(self._response_message_cls.to_json(r)).strip('"') for r in responses
```
?

I'm on the fence, keep it the way it is if you think it's clearer.

tests/unit/test_rest_streaming.py
```diff
@@ -0,0 +1,160 @@
...
+        raise NotImplementedError()
+
+    def iter_content(self, *args, decode_unicode=True, **kwargs):
+        return self._ResponseItr(
+            self._parse_responses(self._responses),
+            decode_unicode,
+            random_split=self._random_split,
+        )
+
+
```
software-dov, line 113: I think that we should parametrize all the tests on `random_split` for increased coverage.

```python
pytest.mark.parametrize("random_split", [True, False]
def test_next_simple(random_split):
```

tests/unit/test_rest_streaming.py
```diff
@@ -0,0 +1,160 @@
...
+import pytest
+import requests
+
+from google.api_core import rest_streaming
+
+
+class Composer(proto.Message):
+    given_name = proto.Field(proto.STRING, number=1)
+    family_name = proto.Field(proto.STRING, number=2)
+
```
software-dov, line 39: Can we also add a `RepeatedField` and a `MapField` for additional testing safety?

google/api_core/rest_streaming.py
software-dov: This is very clever, but it is a little difficult to read. I'm going to need to do a second pass on it. I am a little concerned that it doesn't do any error checking.

tests/unit/test_rest_streaming.py
software-dov: Timestamp and Duration are special. We describe the fields using the `*_pb2.py` proto message types, and then proto-plus handles them as specific special cases.

e.g.
```python
from google.protobuf import duration_pb2
from google.protobuf import timestamp_pb2

import proto

class Event(proto.Message):
    start_time = proto.Field(timestamp_pb2.Timestamp, number=1)
    duration = proto.Field(duration_pb2.Duration, number=2)

import datetime

e = Event(start_time=datetime.datetime(), duration=datetime.timedelta())
```

google/api_core/rest_streaming.py
software-dov: My main concern is readability: if performance becomes a concern we can do some profiling.

google/api_core/rest_streaming.py
atulep: It ignores commas between JSON objects inside of an array (i.e. level == 1). It appends them to `_obj` within a JSON object (i.e. level == 2 and above) - it's handled in the default `else` statement at the very bottom.

google/api_core/rest_streaming.py
atulep: Yeah, I agree the for-loop is complicated, but it's the least worst option available. I considered other alternatives:

1) Use polymorphism. This removes if-else statements but requires a lot more code and perhaps not so Pythonic here.
2) Create DFA class. This for-loop simulates a finite automata, so I can create a separate FA class and this will hide most of the if-else complexity. But, it requires quite a bit of careful formulation and thought to correctly create states and transitions.
3) What Vadym suggested above (e.g. `dict`). I don't think this improves readability as I either need to nest if-statements or create private functions, which IMO introduces indirection here.

Out of 3 options, the 2nd (DFA) is the best IMO, but it would take me more time to complete, putting our deadlines at risk.

atulep: What kind of error checking are you referring? My assumption is server sends valid JSONs.

tests/unit/test_rest_streaming.py
atulep: Done.

tests/unit/test_rest_streaming.py
atulep: Added.

google/api_core/rest_streaming.py
atulep: Done

tests/unit/test_rest_streaming.py
atulep: Done. mypy complains:
```
nox > mypy google tests
tests/unit/test_rest_streaming.py:31: error: Module has no attribute "time_ns"
Found 1 error in 1 file (checked 82 source files)
```

tests/unit/test_rest_streaming.py
atulep: Done

tests/unit/test_rest_streaming.py
atulep: Done

tests/unit/test_rest_streaming.py
atulep: Done

google/api_core/rest_streaming.py
vam-google: As per discussion today with @software-dov, please make that escaped quotes are processed properly. Please also add unit test for that.

google/api_core/rest_streaming.py
software-dov: > My assumption is server sends valid JSONs
Okay, if we really are guaranteed that, then I'm more sanguine.
Other concern was mentioned in meeting today: iterator ends while we still have unparsed data.

google/api_core/rest_streaming.py
atulep: Done

tests/unit/test_rest_streaming.py
atulep: Done.

google/api_core/rest_streaming.py
```diff
@@ -0,0 +1,122 @@
...
+                self._obj += char
+
+            if char == "\\":
+                # Escaping the "\".
+                if self._next_should_be_escaped:
+                    self._next_should_be_escaped = False
+                else:
+                    self._next_should_be_escaped = True
+            else:
+                self._next_should_be_escaped = False
```
software-dov, line 104: Potential rewrite:
```python
# s/_next_should_be_escaped/_escape_next/

self._escape_next = not self._escape_next if char == "\\" else False
```

tests/unit/test_rest_streaming.py
```diff
@@ -0,0 +1,211 @@
...
+    ]
+    resp = ResponseMock(
+        responses=responses, random_split=random_split, response_cls=Song
+    )
+    itr = rest_streaming.ResponseIterator(resp, Song)
+    assert list(itr) == responses
+
+
+@pytest.mark.parametrize("random_split", [True, False])
+def test_next_escaped_characters_in_string(random_split):
```
software-dov, line 150: Just to be sure, does this test fail when your string escape changes are not present?

software-dov (commented):
Looks good, one minor logic nitpick and one minor TDD nitpick.

tests/unit/test_rest_streaming.py
atulep: Yes, it does.

software-dov: Good, just a quirk of mine to disprove the null hypothesis.

vam-google (approved):
LGTM

software-dov (approved):

google/api_core/rest_streaming.py
atulep: Done.

PR #344, opened 2022-02-25
feat: initial support for Extended Operations

software-dov:
Certain APIs with Long-Running Operations deviate from the semantics
in https://google.aip.dev/151 and instead define custom operation
messages, aka Extended Operations.

This change adds a PollingFuture subclass designed to be used with
Extended Operations. It is analogous and broadly similar to
google.api_core.operation.Operation and subclasses
google.api_core.future.polling.PollingFuture.

The full description of Extended Operation semantics is beyond the
scope of this change.

### DIFF
--- a/google/api_core/extended_operation.py
+++ b/google/api_core/extended_operation.py
@@ -15,7 +15,7 @@
 """Futures for extended long-running operations returned from Google Cloud APIs.
 
 These futures can be used to synchronously wait for the result of a
-lon-running operations using :meth:`ExtendedOperation.result`:
+long-running operations using :meth:`ExtendedOperation.result`:
 
 .. code-block:: python
 
--- a/google/api_core/extended_operation.py
+++ b/google/api_core/extended_operation.py
@@ -132,8 +132,7 @@ class ExtendedOperation(polling.PollingFuture):
                 return
 
             if self.error_code and self.error_message:
-                # TODO(dovs): handle this better.
-                exception = exceptions.from_grpc_status(
+                exception = exceptions.from_http_status(
                     status_code=self.error_code,
                     message=self.error_message,
                     response=self._extended_operation,
--- a/google/api_core/extended_operation.py
+++ b/google/api_core/extended_operation.py
@@ -17,10 +17,22 @@
 These futures can be used to synchronously wait for the result of a
 lon-running operations using :meth:`ExtendedOperation.result`:
 
-.. code-black:: python
+.. code-block:: python
 
     extended_operation = my_api_client.long_running_method()
-    result =
+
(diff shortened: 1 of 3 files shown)


google/api_core/extended_operation.py
```diff
@@ -0,0 +1,175 @@
...
+    def __init__(
+        self, extended_operation, refresh, cancel, retry=polling.DEFAULT_RETRY
+    ):
+        super().__init__(retry=retry)
+        # Note: there is not a concrete type the extended operation must be.
+        # It MUST have fields that correspond to the following, POSSIBLY WITH DIFFERENT NAMES:
+        # * name: str
+        # * status: Union[str, bool, enum.Enum]
+        # * error_code: int
+        # * error_message: str
```
busunkim96, line 60: Can this be added to the class docstring instead so it's visible in generated ref docs?

google/api_core/extended_operation.py
```diff
@@ -0,0 +1,175 @@
...
+        #         option (google.cloud.operation_polling_method) = true;
+        #     }
+        # }
+        #
+        # Any info needed for the poll, e.g. a name, path params, etc.
+        # is held in the request, which the initial client method is in a much
+        # better position to make made because the caller made the initial request.
+        #
+        # TL;DR: the caller sets up closures for refresh and cancel that carry
+        # the properly configured requests.
```
busunkim96, line 174: Could this section be turned into a method docstring? Is the idea that most folks will create an `ExtendedOperation` through this method?

busunkim96 (commented):
This looks good! I left a few questions about docs.

google/api_core/extended_operation.py
```diff
@@ -0,0 +1,175 @@
...
+
+"""Futures for extended long-running operations returned from Google Cloud APIs.
+
+These futures can be used to synchronously wait for the result of a
+lon-running operations using :meth:`ExtendedOperation.result`:
+
+.. code-black:: python
+
+    extended_operation = my_api_client.long_running_method()
+    result =
```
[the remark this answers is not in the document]
software-dov: Oh. No, it is not :P

google/api_core/extended_operation.py
software-dov: Of course.

google/api_core/extended_operation.py
software-dov: Definitely. My intention was to have users instantiate `ExtendedOperation` (or relevant subclass) through this factory so that if more work needs to be done later on the child constructors don't need to do as much work.

busunkim96 (approved):

PR #356, opened 2022-03-11
docs: clarify user agent string in ClientInfo

### LINKED ISSUE #341: Improve docs to clarify that the user_agent string exists in the `x-goog-api-client` header
The docstring [here](https://github.com/googleapis/python-api-core/blob/main/google/api_core/client_info.py#L57) mentions that the `user_agent` string is added as a prefix to the user agent header. This is confusing to users because the actual name of the HTTP header that contains the `user_agent` string is `x-goog-api-client`.

### LINKED ISSUE #15081: Improve docs to clarify that the user_agent string exists in the `x-goog-api-client` header
The docstring [here](https://github.com/googleapis/python-api-core/blob/main/google/api_core/client_info.py#L57) mentions that the `user_agent` string is added as a prefix to the user agent header. This is confusing to users because the actual name of the HTTP header that contains the `user_agent` string is `x-goog-api-client`.

parthea:
Fixes googleapis/google-cloud-python#15081 🦕


google/api_core/client_info.py
```diff
@@ -39,9 +39,9 @@
 class ClientInfo(object):
     """Client information used to generate a user-agent for API calls.
 
-    This user-agent information is sent along with API calls to allow the
-    receiving service to do analytics on which versions of Python and Google
-    libraries are being used.
+    This user-agent information is sent along with API calls via the
+    `x-goog-api-client` header to allow the receiving service to do analytics
```
tswast, line 43: Not strictly accurate. I believe this client info actually does populate `user-agent`. It's the grpc version that populates `x-goog-api-client`.

tswast (dismissed):

google/api_core/client_info.py
parthea: Good catch! I'll update the wording so that users are aware when to expect a `user-agent` header

parthea: Moving forward, only the `x-goog-api-client` will contain the `user-agent` information. We won't be populating `user-agent` as a separate header. I suggest we keep the wording as is.

google/api_core/client_info.py
```diff
@@ -39,9 +39,9 @@
 class ClientInfo(object):
     """Client information used to generate a user-agent for API calls.
 
-    This user-agent information is sent along with API calls to allow the
-    receiving service to do analytics on which versions of Python and Google
-    libraries are being used.
+    This user-agent information is sent along with API calls via the
```
vchudnov-g, line 42: Please modify the top of this comment:

```
Information used to generate a client library info header sent with API calls
```

vchudnov-g, line 42: nit: I would also suggest renaming the variable in the function s/`ua`/`info`/, since `ua` might lead people to assume "User-Agent"

google/api_core/gapic_v1/client_info.py
```diff
@@ -42,7 +42,7 @@ class ClientInfo(client_info.ClientInfo):
             library, generally used if the client library was not generated
             by gapic or if additional functionality was built on top of
             a gapic client library.
-        user_agent (Optional[str]): Prefix to the user agent header. This is
+        user_agent (Optional[str]): Prefix to the `x-goog-api-client` header. This is
```
vchudnov-g, line 45: I would somehow deprecate this parameter: rename this parameter to be `prefix`, and then add a  deprecated parameter `user_agent` that merely assigns its value to `prefix`.

PR #379, opened 2022-05-14
feat: adds support for audience in client_options

atulep:
Adds support for `audience` in client_options.

### DIFF
--- a/google/api_core/client_options.py
+++ b/google/api_core/client_options.py
@@ -85,6 +85,7 @@ class ClientOptions(object):
         credentials_file=None,
         scopes=None,
         api_key=None,
+        audience=None,
     ):
         if client_cert_source and client_encrypted_cert_source:
             raise ValueError(
@@ -99,6 +100,7 @@ class ClientOptions(object):
         self.credentials_file = credentials_file
         self.scopes = scopes
         self.api_key = api_key
+        self.audience = audience
 
     def __repr__(self):
         return "ClientOptions: " + repr(self.__dict__)
--- a/google/api_core/client_options.py
+++ b/google/api_core/client_options.py
@@ -70,7 +70,11 @@ class ClientOptions(object):
         scopes (Optional[Sequence[str]]): OAuth access token override scopes.
         api_key (Optional[str]): Google API key. ``credentials_file`` and
             ``api_key`` are mutually exclusive.
-        audience (Optional[str]): The desired API audience.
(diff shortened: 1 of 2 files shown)


atulep:
@viacheslav-rostovtsev FYI. This is needed to make `audience` configurable in the downstream GAPIC client.

atulep:
@viacheslav-rostovtsev What can be a good documentation for `audience`?

google/api_core/client_options.py
```diff
@@ -70,6 +70,7 @@ class ClientOptions(object):
         scopes (Optional[Sequence[str]]): OAuth access token override scopes.
         api_key (Optional[str]): Google API key. ``credentials_file`` and
             ``api_key`` are mutually exclusive.
+        audience (Optional[str]): The desired API audience.
```
parthea, line 73: Please could you include an example string to show the correct format?

parthea (approved):
Added one observation

viacheslav-rostovtsev:
- let's call the parameter `api_audience` and not just `audience` because it looks like in the TPC work we might get other type of audience as well.
- The documentation should be something like: 'The intended audience for the API calls to the service that will be set when using certain 3rd party authentication flows. Audience is typically a resource identifier. If not set, the service endpoint value will be used as a default.'
- Good examples of API audience are google service endpoints like `https://language.googleapis.com`

viacheslav-rostovtsev (changes requested):
Let's change this to `api_audience`.

google/api_core/client_options.py
atulep: Done

atulep:
@viacheslav-rostovtsev PTAL again

viacheslav-rostovtsev (approved):
LGTM

google/api_core/client_options.py
```diff
@@ -70,6 +70,11 @@ class ClientOptions(object):
         scopes (Optional[Sequence[str]]): OAuth access token override scopes.
         api_key (Optional[str]): Google API key. ``credentials_file`` and
             ``api_key`` are mutually exclusive.
+        api_audience (Optional[str]): The intended audience for the API calls
+            to the service that will be set when using certain 3rd party
+            authentication flows. Audience is typically a resource identifier.
+            If not set, the service endpoint value will be used as a default.
+            An example of a valid ``api_audience`` is: "https://language.googleapis.com".
```
nicain, line 77: language.googleapis.com is used as an example here.

PR #401, opened 2022-06-10
fix: drop support for grpcio-gcp

parthea:
Closes #398

The `grpcio-gcp` package is not used in any client libraries. It has [not been updated](https://pypi.org/project/grpcio-gcp/#history) since 2018. There is a compatibility issue with `grpcio-gcp` protobuf 4.21.1. This PR proposes removing `grpcio-gcp` since it is not used.

https://github.com/googleapis/python-api-core/blob/e92045b7a34b8d0a6374d6b1f67c1c6095ad33c6/setup.py#L39-L40

PR #418, opened 2022-07-28
fix: restore support for grpcio-gcp

parthea:
This PR restores the feature which was removed in https://github.com/googleapis/python-api-core/pull/401 which is still used in spanner v1 [here](https://github.com/googleapis/python-spanner/blob/v1/setup.py#L32). `grpcio-gcp` is not compatible with protobuf 4.x.x so the feature is only supported with protobuf 4.x.x. This PR also adds a deprecation warning to let users know that support for `grpcio-gcp` will be dropped after January 1, 2024.

docs: add a note that grpcio-gcp is only supported in environments with protobuf < 4.x.x
docs: raise DeprecationWarning when 'grpcio-gcp' is used
fix(deps): require protobuf >= 3.20.1


google/api_core/grpc_helpers.py
```diff
@@ -24,6 +25,27 @@
 import google.auth.credentials
 import google.auth.transport.grpc
 import google.auth.transport.requests
+import google.protobuf
+
+PROTOBUF_VERSION = google.protobuf.__version__
+
+# The grpcio-gcp package only has support for protobuf < 4
+if int(PROTOBUF_VERSION.split(".")[0]) < 4:
```
vam-google, line 33: Minor: guess we can just check if the first two characters are "4." no need to split and convert.

testing/constraints-3.7.txt
```diff
@@ -6,9 +6,10 @@
 # e.g., if setup.py has "foo >= 1.14.0, < 2.0.0dev",
 # Then this file should have foo==1.14.0
 googleapis-common-protos==1.56.2
-protobuf==3.15.0
+protobuf==3.19.0
```
vam-google, line 9: Why not 3.21?

vam-google (approved):
LGTM, just a few minor things

google/api_core/grpc_helpers.py
parthea: addressed in [6225d4f](https://github.com/googleapis/python-api-core/pull/418/commits/6225d4fad49792d88b00584f8c7596c2d02c076c)

testing/constraints-3.7.txt
parthea: I've bumped the version to 3.20.1, the latest 3.x.x version
https://pypi.org/project/protobuf/3.20.1/

PR #438, opened 2022-09-06
feat: merge main branch into v1

parthea:
For google-api-core we did not make breaking changes in the `main` branch beyond removing support for 2.7. This PR ports all of the features from the `main` branch into the `v1` branch.


parthea:
It should be ok to ignore the CLA in this case, as these commits were already merged in main.

vam-google (approved):
LGTM

PR #453, opened 2022-09-26
feat: Add a little bit of typing to google.api_core.retry

asottile-sentry:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-api-core/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea (**N/A**)
- [x] Appropriate docs were updated (if necessary) (**N/A**)


google/api_core/retry.py
```diff
@@ -54,13 +54,14 @@ def check_if_exists():
 
 """
 
-from __future__ import unicode_literals
+from __future__ import annotations
```
asottile-sentry, line 57: this library has `python_requires >= 3.7` so (1) `unicode_literals` does nothing and (2) this is the first version which supports `__future__.annotations`

it might be a good idea to run https://github.com/asottile/pyupgrade on this codebase to remove the remaining python 2isms

asottile-sentry:
:wave: I've rebased this to resolve conflicts (again) -- could it get reviewed?

google/api_core/retry.py
parthea: Thanks! I've captured this TODO in https://github.com/googleapis/python-api-core/issues/524

parthea (dismissed):
Thanks @asottile-sentry!

parthea (approved):

PR #462, opened 2022-10-19
fix: Major refactoring of Polling, Retry and Timeout logic

vam-google:
If you are a  user of python-api-core and experience polling timing out after ~15 minutes (900s) in your python code after this change, please make sure that instead of calling `PollingFuture.result()` you call it with additonal `timeout` argument like this: `PollingFuture.result(timeout = <desired timeout in seconds>)` or `PollingFuture.result(timeout = None)` (for infinite timeout, but infinite timeouts are strongly **discouraged**) for  your code to stop timing out at 900s.

The core libraries cannot and should not have infitine polling as as default behavior. The fact that it had it like that in python for years (unlike other GCP-supported languages) was a bug, as it contradicts original cross-language LRO methods design in GAPIC libraries and in general users are not supposed to be put into an infinite-loop scenario implicitly. If an LRO can actually ran for hours or even days and it is Ok, then users are expected to acknowledge this fact by providing the huge timeouts explicitly.

This is in response to https://freeman.vc/notes/aws-vs-gcp-reliability-is-wildly-different, which triggered an investigation of the whole Polling/Retry/Timeout behavior in Python GAPIC clients and revealed many fundamental flaws in its implementation.

To properly describe the refactoring in this PR we need to stick to a rigorous terminology, as vague definitions of retries, timeouts, polling and related concepts seems to be the main source of the present bugs and overall confusion among both groups: users of the library and creators of the library. Please check the updated (in this PR) documentation of the `google.api_core.retry.Retry` class and the `google.api_core.future.polling.Polling.result()` method for the proper definitions and context.

Note, the overall semantics around Polling, Retry and Timeout remains quite confusing even after refactoring (although it is now more or less rigorously defined), but it was as clean as I could make it while still maintaining backward compatibility of the whole library.

The quick summary of the changes in this PR:

1) Properly define and fix the application of Deadline and Timeout concepts. Please check the updated documentation for the `google.api_core.retry.Retry` class for the actual definitions. Originally the `deadline` has been used to represent timeouts conflating the two concepts. As result this PR replaces `deadline` arguments with `timeout` ones in as backward-compatible manner as possible (i.e. backward compatible in all practical applications).

2) Properly define RPC Timeout, Retry Timeout and Polling Timeout and how a generic Timeout concept (aka Logical Timeout) is mapped to one of those depending on the context. Please check `google.api_core.retry.Retry` class documentation for details.

3) Properly define and fix the application of Retry and Polling concepts. Please check the updated documentation for `google.api_core.future.polling.PollingFuture.result()` for details.

4) Separate `retry` and `polling` configurations for Polling future, as these are two different concepts (although both operating on `Retry` class). Originally both retry and polling configurations were controlled by a single `retry` parameter, merging configuration regarding how "rpc error responses" and how "operation not completed" responses are supposed to be handled.

5) For the following config properties - `Retry` (including `Retry Timeout`), `Polling` (including `Polling Timeout`) and `RPC Timeout` - fix and properly define how each of the above properties gets configured and which config gets precedence in case of a conflict (check `PollingFuture.result()` method documentation for details). Each of those properties can be specified as follows: directly provided by the user for each call, specified during gapic generation time from config values in `grpc_service_config.json` file (for Retry and RPC Timeout) and `gapic.yaml` file (for Polling Timeout), or be provided as a hard-coded basic default values in python-api-core library itself. This alo includes fixing the per-call polling config propagation logic (the polling/retry configs supplied to `PollingFuture.result()` used to be ignored for actual call).

6) Deprecate `ExponentialTimeout`, `ConstantTimeout` and related logic as those are outdated concepts and are not consistent with the other GAPIC Languages. Replace it with `TimeToDeadlineTimeout` to be consistent with how the rest of the languages do it.

7) Deprecate `google.api_core.operations_v1.config` as it is an outdated concept and self-inconsistent (as all gapic clients provide configuraiton in code). The configs are directly provided in code instead.

8) Switch randomized delay calculation from `delay` being treated as expected value for randomized_delay to `delay` being treated as maximum value for `randomized_delay` (i.e. the new expected valud for `randomized_delay` is `delay / 2`). See the `exponential_sleep_generator()` method implementation for details. This is needed to make Python implementation of retries and polling exponential backoff consistent with the rest of GAPIC languages. Also fix the uncontrollable growth of `delay` value (since it is a subject of exponential growth, the `delay` value was quickly reaching "infinity" value, and the whole thing was not failing simply due to python being a very forgiving language which forgives multiplying "infinity" by a number (`inf * number = inf`) binstead of simply overflowing to a (most likely) negative number). Also essentially rollback  the https://github.com/googleapis/python-api-core/commit/52f12af027568ddeae7fb6b1e3d2111cce1bac08 change, since that is inconsistent with the other languages and damages uniform distibution of retry delays artificially shifting  their concentration towards the end of timeout.

9) Fix url construction in `OperationsRestTransport`. Without this fix the polling logic for REST transport was completely broken (is not affecting Compute client, as that one has custom LRO).

10) Last but not least: change the default values for Polling logic to be the following: `initial=1.0` (same as before), `maximum=20.0` (was `60`), `multiplier=1.5` (was `2.0`), `timeout=900` (was `120`, but due to timeout resolution logic was actually None (i.e. infinity)). This, in conjunction with changed calculation of randomized delay (i.e. its expected value now being  `delay / 2`) overall makes polling logic much less aggressive in terms of increasing delays between each polling iteration, making LRO return much earlier for users on average, but still keeping a healthy balance between strain put on both client and server by polling and responsiveness of LROs for user.

*The design doc summarising all the changes and reasons for them is in progress.

In addition to the timeout/retry fixes, this PR has some other non-related technical fixes: 
- Fix and improve code coverage (and explicitly disables lines which are not supposed to be covered) 
- Define `Python 3.10` as the default python version used in CI.
- Upgrade to Sphinx 4.2.0, as it is the one supporting Python 3.10

### DIFF
--- a/google/api_core/extended_operation.py
+++ b/google/api_core/extended_operation.py
@@ -50,10 +50,13 @@ class ExtendedOperation(polling.PollingFuture):
         refresh (Callable[[], type(extended_operation)]): A callable that returns
             the latest state of the operation.
         cancel (Callable[[], None]): A callable that tries to cancel the operation.
-        retry: Optional(google.api_core.retry.Retry): The retry configuration used
-            when polling. This can be used to control how often :meth:`done`
-            is polled. Regardless of the retry's ``deadline``, it will be
-            overridden by the ``timeout`` argument to :meth:`result`.
+        polling Optional(google.api_core.retry.Retry): The configuration used
+            for polling. This can be used to control how often :meth:`done`
+            is polled. If the ``timeout`` argument to :meth:`result` is
+            specified it will override the ``polling.timeout`` property.
+        retry Optional(google.api_core.retry.Retry): DEPRECATED use ``polling``
+            instead. If specified it will override ``polling`` parameter to
+            maintain backward compatibility.
 
     Note: Most long-running API methods use google.api_core.operation.Operation
     This class is a wrapper for a subset of methods that use alternative
@@ -68,9 +71,14 @@ class ExtendedOperation(polling.PollingFuture):
     """
 
     def __init__(
-        self, extended_operation, refresh, cancel, retry=polling.DEFAULT_RETRY
+        self,
+        extended_operation,
+        refresh,
+        cancel,
+        polling=polling.DEFAULT_POLLING,
+        **kwargs,
     ):
-        super().__init__(retry=retry)
+        super().__init__(polling=polling, **kwargs)
         self._extended_operation = extended_operation
         self._refresh = refresh
         self._cancel = cancel
@@ -114,7 +122,7 @@ class ExtendedOperation(polling.PollingFuture):
     def __getattr__(self, name):
         return getattr(self._extended_operation, name)
 
-    def done(self, retry=polling.DEFAULT_RETRY):
+    def done(self, retry=None):
         self._refresh_and_update(retry)
         return self._extended_operation.done
 
@@ -137,9 +145,11 @@ class ExtendedOperation(polling.PollingFuture):
         self._refresh_and_update()
         return self._extended_operation.done
 
-    def _refresh_and_update(self, retry=polling.DEFAULT_RETRY):
+    def _refresh_and_update(self, retry=None):
         if not self._extended_operation.done:
-            self._extended_operation = self._refresh(retry=retry)
+            self._extended_operation = (
+                self._refresh(retry=retry) if retry else self._refresh()
+            )
             self._handle_refreshed_operation()
 
     def _handle_refreshed_operation(self):
--- a/google/api_core/future/polling.py
+++ b/google/api_core/future/polling.py
@@ -234,7 +234,7 @@ class PollingFuture(base.Future):
 
         Args:
             timeout (int): (Optional) How long (in seconds) to wait for the
-                operation to complete.
+                operation to complete. If None, wait indefinitely.
             retry (google.api_core.retry.Retry): (Optional) How to retry the
                 polling RPC. This defines ONLY how the polling RPC call is
                 retried (i.e. what to do if the RPC we used for polling returned
--- a/google/api_core/future/polling.py
+++ b/google/api_core/future/polling.py
@@ -18,7 +18,7 @@ import abc
 import concurrent.futures
 
 from google.api_core import exceptions
-from google.api_core import retry
+from google.api_core import retry as retries
 from google.api_core.future import _helpers
 from google.api_core.future import base
 
@@ -29,14 +29,40 @@ class _OperationNotComplete(Exception):
     pass
 
 
-RETRY_PREDICATE = retry.if_exception_type(
+# DEPRECATED as it conflates RPC retry and polling concepts into one.
+# Use POLLING_PREDICATE instead to configure polling.
+RETRY_PREDICATE = retries.if_exception_type(
     _OperationNotComplete,
     exceptions.TooManyRequests,
     exceptions.InternalServerError,
     exceptions.BadGateway,
     exceptions.ServiceUnavailable,
 )
-DEFAULT_RETRY = retry.Retry(predicate=RETRY_PREDICATE)
+
+# DEPRECATED, use DEFAULT_POLLING to configure LRO polling logic. Construct
+# Retry object using its default values as a baseline for any custom retry logic
+# (to not be confused with polling logic).
+DEFAULT_RETRY = retries.Retry(predicate=RETRY_PREDICATE)
+
+# Poling predicate is supposed to poll only on _OperationNotComplete.
+# Any RPC-specific errors (like ServiceUnavailable) will be handled
+# by retry logic (to not be confused with polling logic) which is triggered for
+# every polling RPC independently of polling logic but within its context.
+POLLING_PREDICATE = retries.if_exception_type(
+    _OperationNotComplete,
+)
+
+# Default polling configuration
+DEFAULT_POLLING = retries.Retry(
+    predicate=POLLING_PREDICATE,
+    initial=1.0,
+    maximum=20.0,
+    multiplier=1.5,
+    timeout=900,
+)
+
+# Default value used to distinguish None and Unset
+_DEFAULT_POLLING_VALUE = object()
 
 
 class PollingFuture(base.Future):
@@ -45,21 +71,29 @@ class PollingFuture(base.Future):
     The :meth:`done` method should be implemented by subclasses. The polling
     behavior will repeatedly call ``done`` until it returns True.
 
+    The actuall polling logic is encapsulated in :meth:`result` method, see
+    documentation for that method for details on how polling works.
+
     .. note::
 
         Privacy here is intended to prevent the final class from
         overexposing, not to prevent subclasses from accessing methods.
 
     Args:
-        retry (google.api_core.retry.Retry): The retry configuration used
-            when polling. This can be used to control how often :meth:`done`
-            is polled. Regardless of the retry's ``deadline``, it will be
-            overridden by the ``timeout`` argument to :meth:`result`.
+        polling (google.api_core.retry.Retry): The configuration used for polling.
+            This parameter controls how often :meth:`done` is polled. If the
+            ``timeout`` argument is specified in :meth:`result` method it will
+            override the ``polling.timeout`` property.
+        retry (google.api_core.retry.Retry): DEPRECATED use ``polling`` instead.
+            If set it will override ``polling`` paremeter for backward
+            compatibility.
     """
 
-    def __init__(self, retry=DEFAULT_RETRY):
+    _DEFAULT_VALUE = object()
+
+    def __init__(self, polling=DEFAULT_POLLING, **kwargs):
         super(PollingFuture, self).__init__()
-        self._retry = retry
+        self._polling = kwargs.get("retry", polling)
         self._result = None
         self._exception = None
         self._result_set = False
@@ -69,11 +103,13 @@ class PollingFuture(base.Future):
         self._done_callbacks = []
 
     @abc.abstractmethod
-    def done(self, retry=DEFAULT_RETRY):
+    def done(self, retry=None):
         """Checks to see if the operation is complete.
 
         Args:
-            retry (google.api_core.retry.Retry): (Optional) How to retry the RPC.
+            retry (google.api_core.retry.Retry): (Optional) How to retry the
+            polling RPC (to not be confused with polling configuration, see
+            the documentation for :meth:`result` for details).
 
         Returns:
             bool: True if the operation is complete, False otherwise.
@@ -81,45 +117,134 @@ class PollingFuture(base.Future):
         # pylint: disable=redundant-returns-doc, missing-raises-doc
         raise NotImplementedError()
 
-    def _done_or_raise(self, retry=DEFAULT_RETRY):
+    def _done_or_raise(self, retry=None):
         """Check if the future is done and raise if it's not."""
-        kwargs = {} if retry is DEFAULT_RETRY else {"retry": retry}
-
-        if not self.done(**kwargs):
+        if not self.done(retry=retry):
             raise _OperationNotComplete()
 
     def running(self):
         """True if the operation is currently running."""
         return not self.done()
 
-    def _blocking_poll(self, timeout=None, retry=DEFAULT_RETRY):
-        """Poll and wait for the Future to be resolved.
-
-        Args:
-            timeout (int):
-                How long (in seconds) to wait for the operation to complete.
-                If None, wait indefinitely.
-        """
+    def _blocking_poll(self, timeout=_DEFAULT_VALUE, retry=None, polling=None):
+        """Poll and wait for the Future to be resolved."""
         if self._result_set:
             return
 
-        retry_ = self._retry.with_deadline(timeout)
+        polling = polling or self._polling
+        if timeout is not PollingFuture._DEFAULT_VALUE:
+            polling = polling.with_timeout(timeout)
 
         try:
-            kwargs = {} if retry is DEFAULT_RETRY else {"retry": retry}
-            retry_(self._done_or_raise)(**kwargs)
+            polling(self._done_or_raise)(retry=retry)
         except exceptions.RetryError:
             raise concurrent.futures.TimeoutError(
-                "Operation did not complete within the designated " "timeout."
+                "Operation did not complete within the designated timeout."
             )
 
-    def result(self, timeout=None, retry=DEFAULT_RETRY):
-        """Get the result of the operation, blocking if necessary.
+    def result(self, timeout=_DEFAULT_VALUE, retry=None, polling=None):
+        """Get the result of the operation.
+
+        This method will poll for operation status periodically, blocking if
+        necessary. If you just want to make sure that this method does not block
+        for more than X seconds and you do not care about the nitty-gritty of
+        how this  method operates, just call it with ``result(timeout=X)``. The
+        other parameters are for advanced use only.
+
+        Every call to this method is controlled by the following three
+        parameters each of which has a specific distinct role although all three
+        may look very similar: ``timeout``, ``retry`` and ``polling``. In most
+        cases users do not need to specify any custom values for any of these
+        parameters and rely on default ones instead.
+
+        If you choose to specify your custom parameters, please make sure you've
+        read the documentation below carefully.
+
+        First please check :class:`google.api_core.retry.Retry`
+        class documentation for the proper definition of timeout and deadline
+        terms and for the definition the three different types of timeouts.
+        This class operates in terms of Retry Timeot and Polling Timeout, it
+        does let customizing RPC timeout and a user is expected to rely on
+        default behavior for it.
+
+        The roles of each argument of this method are as follows:
+
+        ``timeout`` (int): (Optional) The Polling Timeout as defined in
+        :class:`google.api_core.retry.Retry`. If the operation does not complete
+        within this timeout an exception will be thrown. This parameter affects
+        neither Retry Timeout nor RPC Timeout.
+
+        ``retry`` (google.api_core.retry.Retry): (Optional) How to retry the
+        polling RPC. The ``retry.timeout`` propery of this parameter is the
+        Retry Timeout as defined in :class:`google.api_core.retry.Retry`.
+        This parameter defines ONLY how the polling RPC call is retried
+        (i.e. what to do if the RPC we used for polling returned an error), it
+        does  NOT define how the polling is done (i.e. how frequently and for
+        how long to call the polling RPC - use ``polling`` parameter for that).
+        If a polling RPC throws and error and retrying it fails, the whole
+        future fails with the corresponding exception. If you want to tune which
+        server response error codes are not fatal for operation polling use this
+        parameter to control that (``retry.predicate`` in particular).
+
+        ``polling`` (google.api_core.retry.Retry): (Optional) How often and
+        for how long to call the polling RPC periodically (i.e. what to do if
+        a polling rpc returned successfully but its returned result indicates
+        that the long running operaiton is not completed yet, so we need to
+        check it again at some point in future). This parameter does NOT define
+        how to retry each individual polling RPC in case of an error (use the
+        ``retry`` parameter for that). The ``polling.timeout`` of this parameter
+        is Polling Timeout as defined in as defined in
+        :class:`google.api_core.retry.Retry`.
+
+        For each of the arguments there are also default values in place, which
+        will be used if a user does not specify their own. The default values
+        for the three parameters are not to be confused with the default values
+        for the corresponding arguments in this method (those serve as "not set"
+        markers for the resoluiton logic).
+
+        If ``timeout`` is provided (i.e.``timeout is not _DEFAULT VALUE``, note
+        the `None` value means "infinite timeout") it will be used to control
+        the actual Polling Timeout. Otherwise, ``polling.timeout`` value
+        will be used instead (see below for how the ``polling`` config itself
+        gets resolved). In other words this parameter  effectively overrides
+        the ``polling.timeout`` value if specified. This is so to preserve
+        backward compatibility.
+
+        If ``retry`` is provided (i.e. ``retry is not None``) it will be used to
+        control retry behavior for the polling RPC and the ``retry.timeout``
+        will determine the Retry Timeout. If not provided, the
+        polling RPC will be called with whichever default retry config was
+        specified for the polling RPC at the moment of the construction of the
+        polling RPC's client. For example, if the polling RPC is
+        `operations_client.get_operation()`  the ``retry`` parameter will be
+        controlling its retry behavior (not polling  behavior) and, if not
+        specified, that specific method (``operations_client.get_operation()``)
+        will be retried according to the default retry config provided during
+        creation of ``operations_client`` client instead. This argument exists
+        mainly for backward compatibility, users are very unlikely to ever need
+        to set this parameter explicitly.
+
+        If ``polling`` is provided (i.e. ``polling is not None``) it will be used
+        to controll the overall polling behavior and ``polling.timeout`` will
+        controll Polling Timeout unless it is overridden by ``timeout`` parameter
+        as described above. If not provided the``polling`` parameter specified
+        during construction of this future (the ``polling`` argument in the
+        constructor) will be used instead. Note, since ``timeout`` argument may
+        override ``polling.timeout`` value, this parameter should be viewed as
+        coupled with the ``timeout`` parameter as described above.
 
         Args:
-            timeout (int):
-                How long (in seconds) to wait for the operation to complete.
-                If None, wait indefinitely.
+            timeout (int): (Optional) How long (in seconds) to wait for the
+                operation to complete.
+            retry (google.api_core.retry.Retry): (Optional) How to retry the
+                polling RPC. This defines ONLY how the polling RPC call is
+                retried (i.e. what to do if the RPC we used for polling returned
+                an error), it does  NOT define how the polling is done (i.e. how
+                frequently and for how long to call the polling RPC).
+            polling (google.api_core.retry.Retry): (Optional) How often and
+                for how long to call polling RPC periodically. This parameter
+                does NOT define how to retry each individual polling RPC call
+                (use the ``retry`` parameter for that).
 
         Returns:
             google.protobuf.Message: The Operation's result.
@@ -128,8 +253,8 @@ class PollingFuture(base.Future):
             google.api_core.GoogleAPICallError: If the operation errors or if
                 the timeout is reached before the operation completes.
         """
-        kwargs = {} if retry is DEFAULT_RETRY else {"retry": retry}
-        self._blocking_poll(timeout=timeout, **kwargs)
+
+        self._blocking_poll(timeout=timeout, retry=retry, polling=polling)
 
         if self._exception is not None:
             # pylint: disable=raising-bad-type
@@ -138,12 +263,18 @@ class PollingFuture(base.Future):
 
         return self._result
 
-    def exception(self, timeout=None):
+    def exception(self, timeout=_DEFAULT_VALUE):
         """Get the exception from the operation, blocking if necessary.
+
+        See the documentation for the :meth:`result` method for details on how
+        this method operates, as both ``result`` and this method rely on the
+        exact same polling logic. The only difference is that this method does
+        not accept ``retry`` and ``polling`` arguments but relies on defaul ones
+        instead.
 
         Args:
             timeout (int): How long to wait for the operation to complete.
-                If None, wait indefinitely.
+            If None, wait indefinitely.
 
         Returns:
             Optional[google.api_core.GoogleAPICallError]: The operation's
--- a/google/api_core/future/polling.py
+++ b/google/api_core/future/polling.py
@@ -55,10 +55,10 @@ POLLING_PREDICATE = retries.if_exception_type(
 # Default polling configuration
 DEFAULT_POLLING = retries.Retry(
     predicate=POLLING_PREDICATE,
-    initial=1.0,
-    maximum=20.0,
+    initial=1.0,  # seconds
+    maximum=20.0,  # seconds
     multiplier=1.5,
-    timeout=900,
+    timeout=900,  # seconds
 )
 
 
@@ -137,7 +137,8 @@ class PollingFuture(base.Future):
             polling(self._done_or_raise)(retry=retry)
         except exceptions.RetryError:
             raise concurrent.futures.TimeoutError(
-                "Operation did not complete within the designated timeout."
+                f"Operation did not complete within the designated timeout of "
+                f"{polling.timeout} seconds."
             )
 
     def result(self, timeout=_DEFAULT_VALUE, retry=None, polling=None):
--- a/google/api_core/future/polling.py
+++ b/google/api_core/future/polling.py
@@ -39,14 +39,14 @@ RETRY_PREDICATE = retries.if_exception_type(
     exceptions.ServiceUnavailable,
 )
 
-# DEPRECATED, use DEFAULT_POLLING to configure LRO polling logic. Construct
+# DEPRECATED: use DEFAULT_POLLING to configure LRO polling logic. Construct
 # Retry object using its default values as a baseline for any custom retry logic
-# (to not be confused with polling logic).
+# (not to be confused with polling logic).
 DEFAULT_RETRY = retries.Retry(predicate=RETRY_PREDICATE)
 
-# Polling predicate is supposed to poll only on _OperationNotComplete.
+# POLLING_PREDICATE is supposed to poll only on _OperationNotComplete.
 # Any RPC-specific errors (like ServiceUnavailable) will be handled
-# by retry logic (to not be confused with polling logic) which is triggered for
+# by retry logic (not to be confused with polling logic) which is triggered for
 # every polling RPC independently of polling logic but within its context.
 POLLING_PREDICATE = retries.if_exception_type(
     _OperationNotComplete,
@@ -68,7 +68,7 @@ class PollingFuture(base.Future):
     The :meth:`done` method should be implemented by subclasses. The polling
     behavior will repeatedly call ``done`` until it returns True.
 
-    The actuall polling logic is encapsulated in :meth:`result` method, see
+    The actuall polling logic is encapsulated in :meth:`result` method. See
     documentation for that method for details on how polling works.
 
     .. note::
@@ -82,7 +82,7 @@ class PollingFuture(base.Future):
             ``timeout`` argument is specified in :meth:`result` method it will
             override the ``polling.timeout`` property.
         retry (google.api_core.retry.Retry): DEPRECATED use ``polling`` instead.
-            If set it will override ``polling`` paremeter for backward
+            If set, it will override ``polling`` paremeter for backward
             compatibility.
     """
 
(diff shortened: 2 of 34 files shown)


google/api_core/future/polling.py
```diff
@@ -29,14 +29,40 @@ class _OperationNotComplete(Exception):
...
     exceptions.ServiceUnavailable,
 )
-DEFAULT_RETRY = retry.Retry(predicate=RETRY_PREDICATE)
+
+# DEPRECATED, use DEFAULT_POLLING to configure LRO polling logic. Construct
+# Retry object using its default values as a baseline for any custom retry logic
+# (to not be confused with polling logic).
+DEFAULT_RETRY = retries.Retry(predicate=RETRY_PREDICATE)
+
+# Poling predicate is supposed to poll only on _OperationNotComplete.
```
atulep, line 47: Polling

google/api_core/extended_operation.py
```diff
@@ -68,9 +71,14 @@ class ExtendedOperation(polling.PollingFuture):
     """
 
     def __init__(
-        self, extended_operation, refresh, cancel, retry=polling.DEFAULT_RETRY
```
atulep, line 71: Right now, setting retry will be no-op. It will be a confusing user behavior. Ideally, we would warn developer somehow. I don't think we handle this in GAPIC right now. Something to consider - https://docs.python.org/3/library/exceptions.html#DeprecationWarning.

google/api_core/extended_operation.py
```diff
@@ -114,7 +122,7 @@ def error_message(self):
     def __getattr__(self, name):
         return getattr(self._extended_operation, name)
 
-    def done(self, retry=polling.DEFAULT_RETRY):
+    def done(self, retry=None):
         self._refresh_and_update(retry)
```
atulep, line 126: What happens if user wants to use a different polling other than DEFAULT_RETRY ?

google/api_core/extended_operation.py
```diff
@@ -137,9 +145,11 @@ def cancelled(self):
         self._refresh_and_update()
         return self._extended_operation.done
 
-    def _refresh_and_update(self, retry=polling.DEFAULT_RETRY):
+    def _refresh_and_update(self, retry=None):
         if not self._extended_operation.done:
-            self._extended_operation = self._refresh(retry=retry)
+            self._extended_operation = (
+                self._refresh(retry=retry) if retry else self._refresh()
```
atulep, line 151: Simplify with self._refresh(retry=retry)

google/api_core/future/polling.py
```diff
@@ -29,14 +29,37 @@ class _OperationNotComplete(Exception):
...
+    _OperationNotComplete,
+)
+
+# Default polling configuration
+DEFAULT_POLLING = retries.Retry(
+    predicate=POLLING_PREDICATE,
+    initial=1.0,
+    maximum=20.0,
+    multiplier=1.5,
+    timeout=900,
```
atulep, line 61: Add comment with units.

google/api_core/future/polling.py
```diff
@@ -69,57 +100,149 @@ def __init__(self, retry=DEFAULT_RETRY):
...
-        Args:
-            timeout (int):
-                How long (in seconds) to wait for the operation to complete.
-                If None, wait indefinitely.
-        """
         if self._result_set:
             return
 
-        retry_ = self._retry.with_deadline(timeout)
+        polling = polling or self._polling
```
atulep, line 132: I wouldn't do it. It's confusing if I don't pass a `polling` or explcitly set it to None, but function will overwrite it. Why don't remove from arg list?

google/api_core/future/polling.py
```diff
@@ -69,57 +100,149 @@ def __init__(self, retry=DEFAULT_RETRY):
...
+            polling = polling.with_timeout(timeout)
 
         try:
-            kwargs = {} if retry is DEFAULT_RETRY else {"retry": retry}
-            retry_(self._done_or_raise)(**kwargs)
+            polling(self._done_or_raise)(retry=retry)
         except exceptions.RetryError:
             raise concurrent.futures.TimeoutError(
-                "Operation did not complete within the designated " "timeout."
+                "Operation did not complete within the designated timeout."
```
atulep, line 140: Can you include timeout value in the exception ?

google/api_core/extended_operation.py
vam-google: if `retry=None` then `self._refresh(retry=retry)` and `self._refresh()` are different things, becuase `_refresh` is  most likely defined as `_refrhesh(retry=DEFAULT)`. Note in the whole semantics of gapics `retry/polling` configs `None` has special meaning, and is different from being "not set" or "default".

google/api_core/future/polling.py
vam-google: added

google/api_core/future/polling.py
vam-google: Because this the contract all gapic libraries have. Plling or retry configs can be set by users explicitly or, if not set, default ones will be applied. This is a cross-langauge concept. As for why polling was even added in the args in this PR: this is to fix bug in a backward compatible way: originally `retry` argument was viewed as the one controlling polling, although it does not do it, so a new polling argument was added to provide that functionality to the users (i.e. to have an ability to config polling per method call).

google/api_core/future/polling.py
vam-google: added

google/api_core/extended_operation.py
vam-google: Setting retry will not be no-op, it will overwrite settng polling for backward compatibility int `PollingFuture.__init__()`:
```
self._polling = kwargs.get("retry", polling)
```

The deprecation warning is not thrown to avoid contaminating users logs, since it will be affecting every current user.

vam-google:
@atulep Addressed your comments PTAL

tests/asyncio/gapic/test_method_async.py
```diff
@@ -198,41 +198,6 @@ async def test_wrap_method_with_overriding_retry_and_timeout(unused_sleep):
...
 
 
-@mock.patch("asyncio.sleep")
-@mock.patch(
-    "google.api_core.datetime_helpers.utcnow",
-    side_effect=_utcnow_monotonic(),
-    autospec=True,
-)
-@pytest.mark.asyncio
-async def test_wrap_method_with_overriding_retry_deadline(utcnow, unused_sleep):
```
atulep, line 208: Why was this removed ?

tests/asyncio/test_retry_async.py
```diff
@@ -277,7 +277,7 @@ async def test___call___and_execute_success(self, sleep):
         sleep.assert_not_called()
 
     # Make uniform return half of its maximum, which is the calculated sleep time.
-    @mock.patch("random.uniform", autospec=True, side_effect=lambda m, n: n / 2.0)
```
atulep, line 280: Why was this removed ?

tests/unit/test_retry.py
```diff
@@ -406,8 +407,17 @@ def increase_time(sleep_delay):
         last_wait = sleep.call_args.args[0]
         total_wait = sum(call_args.args[0] for call_args in sleep.call_args_list)
 
-        assert last_wait == 2.9  # and not 8.0, because the last delay was shortened
-        assert total_wait == 9.9  # the same as the deadline
+        assert last_wait == 8.0
```
atulep, line 410: Why is it 8? i actually don't get

tests/unit/gapic/test_method.py
```diff
@@ -199,35 +178,35 @@ def test_wrap_method_with_overriding_retry_and_timeout(unusued_sleep):
...
-
-    assert result == 42
-    timeout_args = [call[1]["timeout"] for call in method.call_args_list]
-    assert timeout_args == [5.0, 10.0, 20.0, 26.0, 25.0]
-    assert utcnow.call_count == (
-        1
-        + 5  # First to set the deadline.
-        + 5  # One for each min(timeout, maximum, (DEADLINE - NOW).seconds)
-    )
+# @mock.patch("time.sleep")
```
atulep, line 181: Remove

tests/asyncio/gapic/test_method_async.py
vam-google: Just updated the PR description. The last few sentences of the item 8 describe the removal of unnecessary logic, and this test was testing it. From the PR description for your convenience

> Also essentially rollback the https://github.com/googleapis/python-api-core/commit/52f12af027568ddeae7fb6b1e3d2111cce1bac08 change, since that is inconsistent with the other languages and damages uniform distibution of retry delays artificially shifting their concentration towards the end of timeout.

tests/asyncio/test_retry_async.py
vam-google: This is also a result of the change described in the item 8 of the PR description:

> Switch randomized delay calculation from delay being treated as expected value for randomized_delay to delay being treated as maximum value for randomized_delay (i.e. the new expected valud for randomized_delay is delay / 2). See the exponential_sleep_generator() method implementation for details. This is needed to make Python implementation of retries and polling exponential backoff consistent with the rest of GAPIC languages.

tests/unit/test_retry.py
vam-google: Same thing as the other two - all are the effect of the removal of the rounding for the last attempt.

tests/unit/gapic/test_method.py
vam-google: removed

vam-google:
@atulep Addressed PR feedback, PTAL. Also let me know if you have any more questions (tl;dr; the changes you were commenting on in your last review cycle were all affected by removal of this logic: https://github.com/googleapis/python-api-core/commit/52f12af027568ddeae7fb6b1e3d2111cce1bac08#diff-c118188be6311975226123f406396979b3dbb200e36b425824fba2aa2d675fa7R208, which should not have been there in the first place.

google/api_core/retry.py
```diff
@@ -139,15 +139,15 @@ def exponential_sleep_generator(initial, maximum, multiplier=_DEFAULT_DELAY_MULT
     Yields:
         float: successive sleep intervals.
     """
-    delay = initial
+    delay = min(initial, maximum)
```
atulep, line 142: Are they never negative ?

google/api_core/future/polling.py
```diff
@@ -69,57 +100,150 @@ def __init__(self, retry=DEFAULT_RETRY):
...
-        Args:
-            timeout (int):
-                How long (in seconds) to wait for the operation to complete.
-                If None, wait indefinitely.
-        """
         if self._result_set:
             return
 
-        retry_ = self._retry.with_deadline(timeout)
+        polling = polling or self._polling
```
atulep, line 132: Helpful to add a comment of what type Polling is. Doesn't need to be a formal Pytype type, but something to help the reader.

google/api_core/gapic_v1/method.py
```diff
@@ -52,65 +52,26 @@ def _apply_decorators(func, decorators):
...
+    filtered_decorators = filter(_is_not_none_or_false, reversed(decorators))
 
-    for decorator in decorators:
+    for decorator in filtered_decorators:
         func = decorator(func)
 
     return func
 
 
-def _determine_timeout(default_timeout, specified_timeout, retry):
```
atulep, line 63: Correct me here - this is removed because timeout is being passed as an argument to _blocking_poll ?

tests/unit/test_retry.py
```diff
@@ -361,7 +362,7 @@ def test___call___and_execute_retry(self, sleep, uniform):
         assert on_error.call_count == 1
 
     # Make uniform return half of its maximum, which is the calculated sleep time.
```
atulep, line 364: Update the comment here to remove halfing.

google/api_core/future/polling.py
atulep: > so a new polling argument was added to provide that functionality to the users
Users can already configure polling in the constructor. This functionality already exists right ?

atulep:
Looks good overall. Have last few comments.

google/api_core/retry.py
vam-google: Who are they? initial and maximum? Yes, they are. There is still a chance that a user will set the values as negative, but it is very unlikely and could be done only either by mistake or with intention to break it all. We could have validated that the values are not negative but seems like gapic libraries do not perform these kinds of validation (i.e. by convention we don't go as far as validating ranges of all integer params), so not validating it here (this is not changed in this PR and has been like this before)

google/api_core/future/polling.py
vam-google: What do you mean by "type polling"? Which types of polling are there?

google/api_core/gapic_v1/method.py
vam-google: This is because ExponentialTimeout and ConstantTimeout are both deprecated and should not be used. This method was essentially creating/selecting one of the two deprecated timeout implementations, since both of the timeouts are not needed and the method is prefixed with `_` (not supposed to be used publicly) i just removed it.

tests/unit/test_retry.py
vam-google: Good catch! removed

google/api_core/future/polling.py
atulep: Python type of the `polling` variable.

atulep (approved):

google/api_core/future/polling.py
vam-google: In which constructor? Please be more specific. Do you mean PollingFuture constructor? Then yes. This is how it works: PollingFuture has a polling config (`self._polling`) which will be  used if user did not specify polling explicitly when they call `future.result()` (i.e. use default one, when "default" is the one passed in the constructor, if the user did not specify their custom config).

atulep: Okay, this is an anti-pattern then. If you want to configure per method, it shouldn't be in the constructor. However, I see your point that it will be backwards incompatible to remove from constructor, and need to keep per-method config option. No further action needed.

alexander-fenster:
@vam-google notification test

vam-google:
@vam-google test

google/api_core/retry.py
```diff
@@ -223,22 +224,88 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
...
+    an error is returned from the server.
+
+    Polling Timeout - it is similar to Retry Timeout, but defines how long the
+    client side is allowed to call polling rpc repeatedly to check a status of a
+    long running operaiton. Unlike in the retry case, the polling rpc is
+    expected to succed (its errors are supposed to be handled by the retry
+    logic). The decision if a new polling attemtp needs to be made is made
+    not based on the RPC status code but based on the status of the returned
+    status of an operation (i.e. it is higher level concept than the rpc error
+    codes).
```
vchudnov-g, line 287:
```suggestion
    Polling Timeout: defines how long the
    client side is allowed to call the polling RPC repeatedly to check a status of a
    long-running operation. Each polling RPC is
    expected to succeed (its errors are supposed to be handled by the retry
    logic). The decision as to whether a new polling attempt needs to be made is based
    not on the RPC status code but  on the status of the returned
    status of an operation. In other words: we will poll a long-running operation until the operation is done or the polling timeout expires. Each poll will inform us of the status of the operation. The poll consists of an RPC to the server that may itself be retried as per the poll-specific retry settings in case of errors. The operation-level retry settings do NOT apply to polling-RPC retries.
```

I tried to make this description a bit clearer. I think it's better but still needs some word-smithing.

google/api_core/retry.py
```diff
@@ -223,22 +224,88 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
...
+
+    With the actual timeout types being defined above, the client libraries
+    often refer to just Timeout without clarifying which type specifically
+    that is. In that case the actual timeout type (sometimes also refered to as
+    Logical Timeout) can be determined from the context. If it is a unary rpc
+    call (i.e. a regular one) Timeout usually stands for the RPC Timeout (if
+    provided directly as a standaone value) or Retry Timeout (if provided as
```
vchudnov-g, line 294:
```suggestion
    provided directly as a standalone value) or Retry Timeout (if provided as
```

google/api_core/retry.py
```diff
@@ -223,22 +224,88 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
...
+    ``retry.timeout`` property of the unary RPC's retry config). For
+    ``Operation`` or ``PolllingFuture`` in general Timeout stands for
```
vchudnov-g, line 296:
```suggestion
    ``Operation`` or ``PollingFuture`` in general Timeout stands for
```

google/api_core/retry.py
```diff
@@ -223,22 +224,88 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
...
         predicate (Callable[Exception]): A callable that should return ``True``
             if the given exception is retryable.
         initial (float): The minimum amount of time to delay in seconds. This
             must be greater than 0.
         maximum (float): The maximum amount of time to delay in seconds.
         multiplier (float): The multiplier applied to the delay.
-        deadline (float): How long to keep retrying in seconds. The last sleep
-            period is shortened as necessary, so that the last retry runs at
-            ``deadline`` (and not considerably beyond it).
+        timeout (float): How long to keep retrying in seconds.
```
vchudnov-g, line 306:
```suggestion
        timeout (float): How long to keep retrying, in seconds.
```

google/api_core/retry.py
```diff
@@ -223,22 +224,88 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
...
+    ``retry.timeout`` property of the unary RPC's retry config). For
+    ``Operation`` or ``PolllingFuture`` in general Timeout stands for
+    Polling Timeout.
+
     Args:
         predicate (Callable[Exception]): A callable that should return ``True``
             if the given exception is retryable.
         initial (float): The minimum amount of time to delay in seconds. This
             must be greater than 0.
         maximum (float): The maximum amount of time to delay in seconds.
```
vchudnov-g, line 304:
```suggestion
        maximum (float): The maximum amount of time to delay, in seconds.
```

google/api_core/retry.py
```diff
@@ -223,22 +224,88 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
...
         multiplier (float): The multiplier applied to the delay.
-        deadline (float): How long to keep retrying in seconds. The last sleep
-            period is shortened as necessary, so that the last retry runs at
-            ``deadline`` (and not considerably beyond it).
+        timeout (float): How long to keep retrying in seconds.
+        deadline (float): DEPRECATED: use `timeout` instead. For backward
+            compatibility, if specified it will override ``timeout`` parameter.
```
vchudnov-g, line 308:
```suggestion
            compatibility, if specified it will override the ``timeout`` parameter.
```

google/api_core/retry.py
```diff
@@ -284,31 +353,53 @@ def retry_wrapped_func(*args, **kwargs):
...
                 on_error=on_error,
             )
 
         return retry_wrapped_func
 
     @property
     def deadline(self):
-        return self._deadline
+        """
+        DEPRECATED: use ``timeout`` instead.  Check the ``Retry`` class
```
vchudnov-g, line 365:
```suggestion
        DEPRECATED: use ``timeout`` instead.  Refer to the ``Retry`` class
```

google/api_core/retry.py
```diff
@@ -284,31 +353,53 @@ def retry_wrapped_func(*args, **kwargs):
...
+
+    @property
+    def timeout(self):
+        return self._timeout
 
     def with_deadline(self, deadline):
-        """Return a copy of this retry with the given deadline.
+        """Return a copy of this retry with the given timeout.
+
+        DEPRECATED use :meth:`with_timeout` instead. Check the ``Retry`` class
```
vchudnov-g, line 377:
```suggestion
        DEPRECATED: use :meth:`with_timeout` instead. Refer to the ``Retry`` class
```

google/api_core/retry.py
```diff
@@ -284,31 +353,53 @@ def retry_wrapped_func(*args, **kwargs):
...
+            Retry: A new retry instance with the given timeout.
+        """
+        return self.with_timeout(timeout=deadline)
+
+    def with_timeout(self, timeout):
+        """Return a copy of this retry with the given timeout.
 
         Args:
-            deadline (float): How long to keep retrying.
+            timeout (float): How long to keep retrying in seconds.
```
vchudnov-g, line 392:
```suggestion
            timeout (float): How long to keep retrying, in seconds.
```

google/api_core/retry_async.py
```diff
@@ -68,9 +68,12 @@ async def check_if_exists():
 _DEFAULT_MAXIMUM_DELAY = 60.0  # seconds
 _DEFAULT_DELAY_MULTIPLIER = 2.0
 _DEFAULT_DEADLINE = 60.0 * 2.0  # seconds
+_DEFAULT_TIMEOUT = 60.0 * 2.0  # seconds
 
 
-async def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
+async def retry_target(
```
vchudnov-g, line 74: nit: any way to make the formatting nicer?

google/api_core/retry_async.py
```diff
@@ -84,12 +87,12 @@ async def retry_target(target, predicate, sleep_generator, deadline, on_error=No
             It should return True to retry or False otherwise.
         sleep_generator (Iterable[float]): An infinite iterator that determines
             how long to sleep between retries.
-        deadline (float): How long to keep retrying the target. The last sleep
-            period is shortened as necessary, so that the last retry runs at
-            ``deadline`` (and not considerably beyond it).
+        timeout (float): How long to keep retrying the target.
```
vchudnov-g, line 90:
```suggestion
        timeout (float): How long to keep retrying the target, in seconds.
```

google/api_core/retry_async.py
```diff
@@ -84,12 +87,12 @@ async def retry_target(target, predicate, sleep_generator, deadline, on_error=No
...
         on_error (Callable[Exception]): A function to call while processing a
             retryable exception.  Any error raised by this function will *not*
             be caught.
+        deadline (float): DEPRECATED use ``timeout`` instead. For beckward
+        compatibility, if set it will override ``timeout`` parameter.
```
vchudnov-g, line 95:
```suggestion
        deadline (float): DEPRECATED use ``timeout`` instead. For backward
        compatibility, if set it will override the ``timeout`` parameter.
```

google/api_core/retry_async.py
```diff
@@ -228,28 +236,40 @@ def _replace(
...
             AsyncRetry: A new retry instance with the given deadline.
         """
-        return self._replace(deadline=deadline)
+        return self._replace(timeout=deadline)
+
+    def with_timeout(self, timeout):
+        """Return a copy of this retry with the given timeout.
+
+        Args:
+            timeout (float): How long to keep retrying.
```
vchudnov-g, line 267:
```suggestion
            timeout (float): How long to keep retrying, in seconds.
```

google/api_core/retry_async.py
```diff
@@ -228,28 +236,40 @@ def _replace(
...
             maximum=maximum or self._maximum,
             multiplier=multiplier or self._multiplier,
-            deadline=deadline or self._deadline,
+            timeout=timeout or self._timeout,
             on_error=on_error or self._on_error,
         )
 
     def with_deadline(self, deadline):
         """Return a copy of this retry with the given deadline.
+        DEPRECATED use :meth:`with_timeout` instead.
```
vchudnov-g, line 253:
```suggestion
        DEPRECATED: use :meth:`with_timeout` instead.
```

google/api_core/timeout.py
```diff
@@ -15,7 +15,8 @@
 """Decorators for applying timeout arguments to functions.
 
 These decorators are used to wrap API methods to apply either a constant
-or exponential timeout argument.
+(DEPRECATED), exponential (DEPRECATED) or Deadline-dependent (recommended)
```
vchudnov-g, line 18: put the recommended one first:
"to apply a Deadline-dependent (RECOMMENDED), constant (DEPRECATED), or exponential (DEPRECATED) "

google/api_core/timeout.py
```diff
@@ -66,9 +67,69 @@ def is_thing_ready(timeout=None):
...
+        return func_with_timeout
+
+    def __str__(self):
+        return "<TimeToDeadlineTimeout timeout={:.1f}>".format(self._timeout)
+
+
 class ConstantTimeout(object):
     """A decorator that adds a constant timeout argument.
 
+    DEPRECATED: use `TimeToDeadlineTimeout` instead.
```
vchudnov-g, line 131:
```suggestion
    DEPRECATED: use ``TimeToDeadlineTimeout`` instead.
```

google/api_core/timeout.py
```diff
@@ -140,6 +201,9 @@ def _exponential_timeout_generator(initial, maximum, multiplier, deadline):
 class ExponentialTimeout(object):
     """A decorator that adds an exponentially increasing timeout argument.
 
+    DEPRECATED: the concept of incrementing timeout exponentially has been
+    deprecated. Use `TimeToDeadlineTimeout` instead.
```
vchudnov-g, line 205:
```suggestion
    deprecated. Use ``TimeToDeadlineTimeout`` instead.
```

tests/unit/test_general_helpers.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2022, Google LLC All rights reserved.
```
vchudnov-g, line 1: Let's delete this file if it's not doing anything.

vchudnov-g (commented):
I have not looked carefully at the tests yet, but so far this looks great. I will be running some manual tests to complement what Tony and Aza have been doing.

Most of the comments are documentation tweaks, so hopefully you can just accept them on GitHub and pull them into your local repo before making any other changes.

tests/unit/test_general_helpers.py
vam-google: It is necesssary for test coverage to do its thing. It is an empty test for an empty file (I don't know the history of that, just trying to fix a very broken build here).

google/api_core/future/polling.py
vam-google: I don't understand what you mean. Every polling config is of type Retry.

google/api_core/retry.py
```diff
@@ -198,19 +201,17 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
...
+        if deadline is not None:
+            next_attempt_time = datetime_helpers.utcnow() + datetime.timedelta(
+                seconds=sleep
+            )
+            if deadline < next_attempt_time:
                 raise exceptions.RetryError(
                     "Deadline of {:.1f}s exceeded while calling target function".format(
-                        deadline
+                        timeout
                     ),
```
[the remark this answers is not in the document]
vam-google: This is all automatically formatted, please do not do formatting suggestions as they would violate formatting validation anyways.

google/api_core/retry.py
```diff
@@ -198,19 +201,17 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
...
-
-        if deadline_datetime is not None:
-            if deadline_datetime <= now:
+        if deadline is not None:
+            next_attempt_time = datetime_helpers.utcnow() + datetime.timedelta(
+                seconds=sleep
+            )
+            if deadline < next_attempt_time:
                 raise exceptions.RetryError(
                     "Deadline of {:.1f}s exceeded while calling target function".format(
```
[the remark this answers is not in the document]
vam-google: No, it is one of the places where it is actually deadline.

google/api_core/retry_async.py
vam-google: Please do not suggest formatting, it is autoformatted

google/api_core/future/polling.py
```diff
@@ -29,14 +29,37 @@ class _OperationNotComplete(Exception):
...
+RETRY_PREDICATE = retries.if_exception_type(
     _OperationNotComplete,
     exceptions.TooManyRequests,
     exceptions.InternalServerError,
     exceptions.BadGateway,
     exceptions.ServiceUnavailable,
 )
-DEFAULT_RETRY = retry.Retry(predicate=RETRY_PREDICATE)
+
+# DEPRECATED, use DEFAULT_POLLING to configure LRO polling logic. Construct
```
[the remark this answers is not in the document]
vam-google: The baseliine part was there before. Just keeping it as is, as the current recommendation is simply to not used that at all for anything.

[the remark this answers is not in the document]
vam-google: As for being explicit on usage of DEFAULT_RETRY, it is deprecated, so it should not be used.

google/api_core/future/polling.py
```diff
@@ -69,57 +100,150 @@ def __init__(self, retry=DEFAULT_RETRY):
...
+        if not self.done(retry=retry):
             raise _OperationNotComplete()
 
     def running(self):
         """True if the operation is currently running."""
         return not self.done()
 
-    def _blocking_poll(self, timeout=None, retry=DEFAULT_RETRY):
-        """Poll and wait for the Future to be resolved.
+    def _blocking_poll(self, timeout=_DEFAULT_VALUE, retry=None, polling=None):
```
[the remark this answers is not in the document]
vam-google: Here it is not a refactoring it is a fix, so it does not intend to be backward-compatible (as old behavior was wrong).
It used to be:
1) retry ignored
2) timeout always None, making actual timeout from config also being ignored.

That is being fixed here:
1) Retry is now appreciated as retry config for status checking rpc
2) Timeout will be None (meaning infinite) only if explicitly specified, otherwise it will be the defatult one or from config file (how that is resovled is described in the internal design doc)

google/api_core/future/polling.py
```diff
@@ -69,57 +100,150 @@ def __init__(self, retry=DEFAULT_RETRY):
...
+        constructor) will be used instead. Note, since ``timeout`` argument may
+        override ``polling.timeout`` value, this parameter should be viewed as
+        coupled with the ``timeout`` parameter as described above.
 
         Args:
-            timeout (int):
-                How long (in seconds) to wait for the operation to complete.
-                If None, wait indefinitely.
+            timeout (int): (Optional) How long (in seconds) to wait for the
+                operation to complete.
```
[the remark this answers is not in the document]
vam-google: Added note about None. As for the rest of behavior it is described int the bigger comment for this method. Note, resolution of actual config values used in runtime is surprisingly complex, so it is impossible to describe that in a comment here (there is a whole section in retry design doc describing it).

google/api_core/future/polling.py
```diff
@@ -128,8 +252,8 @@ def result(self, timeout=None, retry=DEFAULT_RETRY):
             google.api_core.GoogleAPICallError: If the operation errors or if
                 the timeout is reached before the operation completes.
         """
-        kwargs = {} if retry is DEFAULT_RETRY else {"retry": retry}
-        self._blocking_poll(timeout=timeout, **kwargs)
+
+        self._blocking_poll(timeout=timeout, retry=retry, polling=polling)
```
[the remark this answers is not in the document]
vam-google: It is not backward compatible by design, because it is a fix (this whole PR is not a refactoring only it also contains many fixes, this is one of them.

google/api_core/gapic_v1/method.py
```diff
@@ -52,65 +52,26 @@ def _apply_decorators(func, decorators):
...
         retry (google.api_core.retry.Retry): The default retry for the
             callable. If ``None``, this callable will not retry by default
-        timeout (google.api_core.timeout.Timeout): The default timeout
-            for the callable. If ``None``, this callable will not specify
-            a timeout argument to the low-level RPC method by default.
+        timeout (google.api_core.timeout.Timeout): The default timeout for the
+            callable (i.e. duration of time within which an RPC must terminate
+            after its start, to not be confused with deadline). If ``None``,
+            this callable will not specify a timeout argument to the low-level
+            RPC method by default.
```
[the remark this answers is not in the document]
vam-google: The "by default" was just redundant, removed it.

google/api_core/operations_v1/transports/rest.py
```diff
@@ -73,6 +74,7 @@ def __init__(
         always_use_jwt_access: Optional[bool] = False,
         url_scheme: str = "https",
         http_options: Optional[Dict] = None,
+        uri_prefix: str = "v1",
```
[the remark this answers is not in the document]
vam-google: Ok, thought it will require a change in gapic-generator-python PR as well

google/api_core/retry.py
```diff
@@ -161,12 +161,14 @@ def retry_target(target, predicate, sleep_generator, deadline, on_error=None):
             It should return True to retry or False otherwise.
         sleep_generator (Iterable[float]): An infinite iterator that determines
             how long to sleep between retries.
-        deadline (float): How long to keep retrying the target. The last sleep
+        timeout (float): How long to keep retrying the target. The last sleep
```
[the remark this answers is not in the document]
vam-google: good catch. The whole "shortening" logic has been removed, so I removed the part of the comment talking about it (including the reference to deadline)

google/api_core/timeout.py
vam-google: Flipped the order

parthea (approved):

google/api_core/future/polling.py
```diff
@@ -69,57 +100,150 @@ def __init__(self, retry=DEFAULT_RETRY):
...
         if self._result_set:
             return
 
-        retry_ = self._retry.with_deadline(timeout)
+        polling = polling or self._polling
+        if timeout is not PollingFuture._DEFAULT_VALUE:
+            polling = polling.with_timeout(timeout)
 
         try:
-            kwargs = {} if retry is DEFAULT_RETRY else {"retry": retry}
```
tswast, line 109: FYI: this was a breaking change. https://github.com/googleapis/python-api-core/issues/477

sasha-gitg: Also for python-aiplatform: https://github.com/googleapis/python-aiplatform/issues/1870

vam-google: that rety logic line never worked (the retry had been i. What broke you is most likely the new default timeout value (instead of None).

jacobg:
It seems like this merge broke using freezegun, particularly with google-cloud-datastore in unit testing. Here's the stack trace:
```
  File "/Users/jacob/.pyenv/versions/gae2/lib/python3.9/site-packages/google/cloud/datastore/transaction.py", line 231, in begin
    response_pb = self._client._datastore_api.begin_transaction(
  File "/Users/jacob/.pyenv/versions/gae2/lib/python3.9/site-packages/google/cloud/datastore_v1/services/datastore/client.py", line 834, in begin_transaction
    response = rpc(
  File "/Users/jacob/.pyenv/versions/gae2/lib/python3.9/site-packages/google/api_core/gapic_v1/method.py", line 113, in __call__
    return wrapped_func(*args, **kwargs)
  File "/Users/jacob/.pyenv/versions/gae2/lib/python3.9/site-packages/google/api_core/timeout.py", line 120, in func_with_timeout
    return func(*args, **kwargs)
  File "/Users/jacob/.pyenv/versions/gae2/lib/python3.9/site-packages/google/api_core/grpc_helpers.py", line 74, in error_remapped_callable
    raise exceptions.from_grpc_error(exc) from exc
google.api_core.exceptions.DeadlineExceeded: 504 Deadline Exceeded
```

Would you expect that to happen? Any workarounds?

PR #489, opened 2023-03-07
fix: remove trailing underscore in transcoding

### LINKED ISSUE #15074: Transcoding fails when HTTP request body contains a field name which is a reserved word in python
There is an error in https://github.com/googleapis/python-debugger-client/pull/151 which occurs in the transcoding logic when the name of the request field whose value is mapped to the HTTP request body is a reserved word in python. In the case of `python-debugger-client`, the `breakpoint` field is used in the HTTP request body [here](https://github.com/googleapis/googleapis/blob/0f35802cb7b4bd752c4adaa9dbdcaaa1ae6bd40a/google/devtools/clouddebugger/v2/debugger.proto#L56).

See the `transcode` function below.
https://github.com/googleapis/python-api-core/blob/2a6a36bd820e1344c8b2526eb75d0b9f3a055d3b/google/api_core/path_template.py#L250-L252

Gapic generator python adds a trailing underscore to field names if the name is also a reserved word in Python. A similar issue was reported in https://github.com/googleapis/python-api-core/issues/227 and fixed in https://github.com/googleapis/python-api-core/pull/228. Based on comment https://github.com/googleapis/python-api-core/issues/227#issuecomment-880273391 it is not possible for fields to have a trailing `_` at the proto level (Creating a field like this generates an error from the protobuf runtime).

See build log https://github.com/googleapis/python-debugger-client/actions/runs/4292539515/jobs/7479094983 which shows the failure when a request has body with `breakpoint_` instead of `breakpoint`.

```
>           response = client.set_breakpoint(request)

tests/unit/gapic/debugger_v2/test_debugger2.py:1920: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
google/cloud/debugger_v2/services/debugger2/client.py:547: in set_breakpoint
    response = rpc(
.nox/unit-3-8/lib/python3.8/site-packages/google/api_core/gapic_v1/method.py:113: in __call__
    return wrapped_func(*args, **kwargs)
.nox/unit-3-8/lib/python3.8/site-packages/google/api_core/timeout.py:120: in func_with_timeout
    return func(*args, **kwargs)
.nox/unit-3-8/lib/python3.8/site-packages/google/api_core/grpc_helpers.py:72: in error_remapped_callable
    return callable_(*args, **kwargs)
google/cloud/debugger_v2/services/debugger2/transports/rest.py:722: in __call__
    transcoded_request = path_template.transcode(http_options, pb_request)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

http_options = [{'body': 'breakpoint', 'method': 'post', 'uri': '/v2/debugger/debuggees/{debuggee_id}/breakpoints/set'}]
message = debuggee_id: "sample1"
breakpoint_ {
  id: "id_value"
  action: LOG
  location {
    path: "path_value"
    line: 424
...alue"
      type_: "type__value"
      var_table_index {
        value: 541
      }
      status {
      }
    }
  }
}
```

parthea:
Fixes https://github.com/googleapis/google-cloud-python/issues/15074
Googlers see proposal go/python-transcoding-reserved-keywords

This PR fixes an error seen in https://github.com/googleapis/python-debugger-client/pull/151 which happens when the gapic generator appends an underscore to a field name which is a python reserved keyword. As similar issue was reported in https://github.com/googleapis/python-api-core/issues/227 and fixed in https://github.com/googleapis/python-api-core/pull/228.  Based on comment https://github.com/googleapis/python-api-core/issues/227#issuecomment-880273391 it is not possible for fields to have trailing `_` at the proto level (Creating a field like this generates an error from the protobuf runtime).

See build log https://github.com/googleapis/python-debugger-client/actions/runs/4292539515/jobs/7479094983 which shows the failure when a request has body with `breakpoint_` instead of `breakpoint_`.


tests/unit/test_path_template.py
```diff
@@ -413,6 +424,25 @@ def test_transcode_with_wildcard(
...
+        [
+            [["post", "/v1/no/template", "data"]],
+            None,
+            {"data": {"id": 1, "info": "some info"}, "foo_": "bar"},
+            ["post", "/v1/no/template", {"id": 1, "info": "some info"}, {"foo": "bar"}],
+        ],
+        # Single field body with reserved keyword, using message where field name has trailing underscore
+        [
+            [["post", "/v1/no/template", "breakpoint"]],
+            SomeMessage(breakpoint_=Breakpoint(name="test"), debuggee_id="test")._pb,
```
vchudnov-g, line 437: minor: could we use distinct values here, say `"test"` and `"exam"`, or `"alice"` and `"bob"`?

google/api_core/path_template.py
```diff
@@ -308,9 +308,27 @@ def transcode(http_options, message=None, **request_kwargs):
...
+                            request["body"] = getattr(leftovers, body)
+                            delete_field(leftovers, body)
+                        except AttributeError as e:
+                            # gapic-generator-python appends underscores to field names
+                            # that collide with python keywords.
+                            # `_` is stripped away as it is not possible to
+                            # natively define a field with a trailing underscore in protobuf.
+                            # See related issue
+                            # https://github.com/googleapis/python-api-core/issues/227
+                            if hasattr(leftovers, body + "_"):
```
vchudnov-g, line 321: If we know that the trailing underscore will fail, why do we have to get an exception before we deal with this? Why not take processing similar to that in lines 327 _et seq_ (stripping `"_"` from `leftovers`) and put it above, and possibly get rid of the `try` block? (Side effect would be also stripping underscores from the other fields in `leftover`, which the code here doesn't do, but I guess we want, so the query params don't have underscores?)

vchudnov-g (commented):
Question about simplifying/robustifying processing...

tests/unit/test_path_template.py
parthea: Fixed.

google/api_core/path_template.py
parthea: I haven't found a way to strip the trailing underscore from a protobuf message. We need to use the correct suffix to access the field. I've proposed a new version without the `try` block. PTAL

vchudnov-g (approved):

google/api_core/path_template.py
```diff
@@ -308,27 +317,18 @@ def transcode(http_options, message=None, **request_kwargs):
...
+                        # `_` is stripped away as it is not possible to
+                        # natively define a field with a trailing underscore in protobuf.
+                        # See related issue
+                        # https://github.com/googleapis/python-api-core/issues/227
+                        field_suffix = ""
+                        if hasattr(leftovers, body + "_"):
+                            field_suffix = "_"
+                        request["body"] = getattr(leftovers, f"{body}{field_suffix}")
+                        delete_field(leftovers, f"{body}{field_suffix}")
+                    else:
```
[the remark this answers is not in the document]
parthea: Correct. This is my understanding also.

google/api_core/path_template.py
```diff
@@ -308,27 +317,18 @@ def transcode(http_options, message=None, **request_kwargs):
...
-                            else:
-                                raise e
-                    else:
-                        # gapic-generator-python appends underscores to field names
+                        # gapic-generator-python append an underscores to field names
                         # that collide with python keywords.
-                        leftovers = {
-                            key.rstrip("_"): val for key, val in leftovers.items()
-                        }
+                        # `_` is stripped away as it is not possible to
```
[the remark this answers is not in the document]
parthea: Done.

google/api_core/path_template.py
```diff
@@ -298,6 +298,15 @@ def transcode(http_options, message=None, **request_kwargs):
         # Assign body and query params
         body = http_option.get("body")
 
+        # gapic-generator-python appends an underscore to field names
+        # that collide with python keywords.
+        # `_` is stripped away as it is not possible to
+        # natively define a field with a trailing underscore in protobuf.
+        # See related issue
+        # https://github.com/googleapis/python-api-core/issues/227
+        if isinstance(leftovers, dict):
```
atulep, line 307: `leftovers` can also be a `proto.Message`, as the original `transcoded_value` ([code](https://github.com/googleapis/python-api-core/blob/main/google/api_core/path_template.py#L274)). 

Shouldn't we take care of it too?

atulep: Oh I see, you're handling it in lines 320-328. I think you can move this logic up here. `leftovers` is either a message or `dict`, so you can do this preprocessing together. It will be more readable too.

parthea: Fixed in [5504d0a](https://github.com/googleapis/python-api-core/pull/489/commits/5504d0ad2112f328d42af3eba5ed3a0a98c2a292)

google/api_core/path_template.py
```diff
@@ -298,6 +298,26 @@ def transcode(http_options, message=None, **request_kwargs):
...
+        # When `leftovers` is a dict, the `_` suffix in each key
+        # is stripped away as it is not possible to natively define a field
+        # with a trailing underscore in protobuf.
+        # When `leftovers` is a protobuf message, we need to use an underscore
+        # suffix when accessing the field in the protobuf message when the
+        # field has an underscore suffix.
+        field_suffix = ""
+
+        if isinstance(leftovers, dict):
+            leftovers = {key.rstrip("_"): val for key, val in leftovers.items()}
```
BenRKarl, line 316: Possible corner case here - are there instances where we should _not_ remove the trailing underscore from a field name? I'm not sure if a protobuf field name can end with an underscore and still be valid, but it might be worth adding some validation to ensure we're only stripping the suffix from the right fields.

parthea: Based on comment https://github.com/googleapis/python-api-core/issues/227#issuecomment-880273391 it is not possible for fields to have trailing `_` at the proto level (Creating a field like this generates an error from the protobuf runtime).

BenRKarl: SGTM thanks for double-checking!

BenRKarl (approved):

PR #492, opened 2023-03-13
fix: fix test_repr assertion

### LINKED ISSUE #449: assert in test_repr is always true
https://github.com/googleapis/python-api-core/blob/c97c4980125a86f384cdf12720df7bb1a2adf9d2/tests/unit/test_client_options.py#L149-L153

Do you mean ` or repr(options) == ...` here?
Because this assert is always true.

vchudnov-g:
[python-api-core/tests/unit/test_client_options.py](https://github.com/googleapis/python-api-core/blob/c97c4980125a86f384cdf12720df7bb1a2adf9d2/tests/unit/test_client_options.py#L149-L153):`test_repr()` had an assertion that was always `True`.

Fixes #449


vchudnov-g:
Stepping back: the intent appears to be to compare regardless of order, so it's more robust and cleaner to restructure the test to not have branches for possible orderings.

bcoe:
> Stepping back: the intent appears to be to compare regardless of order, so it's more robust and cleaner to restructure the test to not have branches for possible orderings.

I looked back at when this test was added, and read through the codebase better (based on your most recent feedback on my PR). Really, the goal of the test is to make sure that:

```python
    def __repr__(self):
        return "ClientOptions: " + repr(self.__dict__)
```

Serializes the class with a `ClientOptions` prefix, followed by a string representation of the keys. It's overkill IMO to assert that every key is set, let's just switch to a simple regex:

https://github.com/googleapis/python-api-core/pull/497

bcoe:
Fixed in https://github.com/googleapis/python-api-core/issues/449, https://github.com/googleapis/python-api-core/pull/492, thanks for your help with review.

PR #495, opened 2023-04-07
feat: retry and retry_async support streaming rpcs

### LINKED ISSUE #485: Retry decorator should support generator functions
**Is your feature request related to a problem? Please describe.**

Server streaming libraries [return an iterable](https://github.com/googleapis/python-bigtable/blob/a6cfb021e1b158b81879b115868112875b1e4162/google/cloud/bigtable_v2/services/bigtable/client.py#L471) that can asynchronously yield data from the backend over time. Some of our libraries need to provide a wrapper around the raw stream to do some local processing, before passing the data to the user.

It would be useful to wrap this whole pipeline in a [retry decorator](https://github.com/googleapis/python-api-core/blob/b8e75cff8bf6772ab76ca6f8d3c80dc2580075c7/google/api_core/retry.py#L222), so that if the stream breaks mid-way through, we can recover and continue yielding data through our generator as if nothing happened. 

Unfortunately, the [current implementation ](https://github.com/googleapis/python-api-core/blob/b8e75cff8bf6772ab76ca6f8d3c80dc2580075c7/google/api_core/retry.py#L191) returns the result of the target function directly, so generators will not yield values and exceptions through the retry block

 **Describe the solution you'd like**

Add some code to detect when a target function is a generator. If so, Replace: https://github.com/googleapis/python-api-core/blob/b8e75cff8bf6772ab76ca6f8d3c80dc2580075c7/google/api_core/retry.py#L191
with:
```
yield from target()
```

Some extra work will also be needed to detect generator `.close()`, and async generators

I plan on putting together a PR for this shortly

daniel-sanche:
From https://github.com/googleapis/python-api-core/issues/485:
> Server streaming libraries [return an iterable](https://github.com/googleapis/python-bigtable/blob/a6cfb021e1b158b81879b115868112875b1e4162/google/cloud/bigtable_v2/services/bigtable/client.py#L471) that can asynchronously yield data from the backend over time. Some of our libraries need to provide a wrapper around the raw stream to do some local processing, before passing the data to the user.
> 
> It would be useful to wrap this whole pipeline in a [retry decorator](https://github.com/googleapis/python-api-core/blob/b8e75cff8bf6772ab76ca6f8d3c80dc2580075c7/google/api_core/retry.py#L222), so that if the stream breaks mid-way through, we can recover and continue yielding data through our generator as if nothing happened.
> 
> Unfortunately, the [current implementation ](https://github.com/googleapis/python-api-core/blob/b8e75cff8bf6772ab76ca6f8d3c80dc2580075c7/google/api_core/retry.py#L191)returns the result of the target function directly, so generators will not yield values and exceptions through the retry block

This PR addresses the issue by adding `retry_target_generator` functions to both the async and sync retry modules, which yield through the target rather than call it directly. Generator mode can be enabled using the `is_generator` argument on the decorator.

Fixes https://github.com/googleapis/python-api-core/issues/485


google/api_core/retry.py
```diff
@@ -331,10 +415,11 @@ def __call__(self, func, on_error=None):
             on_error (Callable[Exception]): A function to call while processing
                 a retryable exception. Any error raised by this function will
                 *not* be caught.
-
+                If `is_generator` is set, non-None values returned by
+                `on_error` will be yielded for downstream consumers.
         Returns:
             Callable: A callable that will invoke ``func`` with retry
-                behavior.
+                behavior``.
```
dizcology, line 422: What are the backticks here for?

google/api_core/retry.py
```diff
@@ -145,6 +145,85 @@ def exponential_sleep_generator(initial, maximum, multiplier=_DEFAULT_DELAY_MULT
...
+        deadline = None
+
+    last_exc = None
+    subgenerator = None
+    for sleep in sleep_generator:
+        # Start a new retry loop
+        try:
+            # create and yeild from a new instance of the generator from input generator function
+            subgenerator = target()
+            return (yield from subgenerator)
```
dizcology, line 195: Between `retry_target` and `retry_target_generator`, what are some differences besides this?  If all the surrounding behaviors are identical, is this maybe a chance to refactor the code somewhat?

(I have the same questions about the async versions of the two functions.)

daniel-sanche: Refactoring like that has more complications than are visible at first glance, because the presence of `yield` automatically changes a function's signature into a generator function, even if it's behind a conditional branch that's not executed. So we need different entrypoint functions for generator vs standard wrappers

I originally tried pulling out some of the shared code blocks at the beginning and end into helper functions, but a lot of context arguments were needed, and it ended up being only a couple lines that could be shared, so it felt less readable than having some duplicated structure.

daniel-sanche: That said, looking at it again, there may be a different approach that could work, by writing a custom Generator class implementation, and returning that instead of relying on the built-in `yield`. I might give that a look to see if that can make things cleaner.

google/api_core/retry_streaming.py
```diff
@@ -0,0 +1,183 @@
...
+        Args:
+            target: The function to call to produce iterables for each retry.
+                This must be a nullary function - apply arguments with
+                `functools.partial`.
+            predicate: A callable used to determine if an
+                exception raised by the target should be considered retryable.
+                It should return True to retry or False otherwise.
+            sleep_generator: An infinite iterator that determines
+                how long to sleep between retries.
+            timeout: How long to keep retrying the target.
```
parthea, line 55: Please specify the units. eg `in seconds`

tests/unit/test_retry.py
```diff
@@ -1,4 +1,3 @@
-# Copyright 2017 Google LLC
```
parthea, line 1: Please preserve the copyright line

tests/unit/test_retry.py
```diff
@@ -162,6 +162,21 @@ def test_retry_target_bad_sleep_generator():
...
+    from google.api_core.retry_streaming import RetryableGenerator
+
+    def target_fn():
+        def inner_gen():
+            raise RuntimeError("initiate retry")
+            yield None
+
+        return inner_gen()
+
+    with pytest.raises(ValueError, match="Sleep generator"):
```
parthea, line 175:
```suggestion
    with pytest.raises(ValueError, match="Sleep generator stopped yielding sleep values"):
```

tests/unit/test_retry.py
```diff
@@ -467,3 +482,261 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
         sleep.assert_any_call(retry_._initial)
+
+    def _generator_mock(
+        self,
+        num=5,
+        error_on=None,
+        return_val=None,
+        exceptions_seen=None,
+        ignore_sent=False,
+    ):
```
parthea, line 493: Please add a docstring explaining the purpose of this function

tests/unit/test_retry.py
```diff
@@ -467,3 +482,261 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+    @mock.patch("time.sleep", autospec=True)
+    def test___call___generator_success(self, sleep):
+        import types
+        import collections
+
+        retry_ = retry.Retry()
+
+        decorated = retry_(self._generator_mock)
+
+        num = 10
```
parthea, line 518: nit: change `num` to be more descriptive. What is it?

tests/unit/test_retry.py
```diff
@@ -467,3 +482,261 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+                    sent_in = None
+            return return_val
+        except (Exception, BaseException, GeneratorExit) as e:
+            # keep track of exceptions seen by generator
+            if exceptions_seen is not None:
+                exceptions_seen.append(e)
+            raise
+
+    @mock.patch("time.sleep", autospec=True)
+    def test___call___generator_success(self, sleep):
```
parthea, line 510: Add a docstring to explain what we're checking for in this test.

tests/unit/test_retry.py
```diff
@@ -467,3 +482,261 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        assert isinstance(self._generator_mock(num), types.GeneratorType)
+        # check yield contents
+        unpacked = [i for i in result]
+        assert len(unpacked) == num
+        for a, b in zip(decorated(num), self._generator_mock(num)):
+            assert a == b
+        sleep.assert_not_called()
+
+    @mock.patch("time.sleep", autospec=True)
+    def test___call___generator_retry(self, sleep):
```
parthea, line 533: Add a docstring to explain what we're checking for in this test.

tests/unit/test_retry.py
```diff
@@ -467,3 +482,261 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        result = retry_(self._generator_mock)(error_on=3)
+        # error thrown on 3
+        # generator should contain 0, 1, 2 looping
+        unpacked = [next(result) for i in range(10)]
+        assert unpacked == [0, 1, 2, 0, 1, 2, 0, 1, 2, 0]
+        assert on_error.call_count == 3
+
+    @mock.patch("random.uniform", autospec=True, side_effect=lambda m, n: n)
+    @mock.patch("time.sleep", autospec=True)
+    def test___call___generator_retry_hitting_deadline(self, sleep, uniform):
```
parthea, line 550: Add a docstring to explain what we're checking for in this test.

tests/unit/test_retry.py
```diff
@@ -467,3 +482,261 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        found_value = None
+        try:
+            while True:
+                next(generator)
+        except StopIteration as e:
+            found_value = e.value
+        assert found_value == expected_value
+
+    @mock.patch("time.sleep", autospec=True)
+    def test___call___with_generator_close(self, sleep):
```
parthea, line 677: Add a docstring to explain what we're checking for in this test.

tests/unit/test_retry.py
```diff
@@ -467,3 +482,261 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        for i in range(2):
+            next(generator)
+        generator.close()
+        assert isinstance(exception_list[0], GeneratorExit)
+        with pytest.raises(StopIteration):
+            # calling next on closed generator should raise error
+            next(generator)
+
+    @mock.patch("time.sleep", autospec=True)
+    def test___call___with_generator_throw(self, sleep):
```
parthea, line 693: Add a docstring to explain what we're checking for in this test.

tests/unit/test_retry.py
```diff
@@ -467,3 +482,261 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        for i in range(2):
+            next(generator)
+        val = generator.throw(ValueError("test"))
+        assert val == 0
+        assert isinstance(exception_list[0], ValueError)
+        # calling next on closed generator should not raise error
+        assert next(generator) == 1
+
+    @mock.patch("time.sleep", autospec=True)
+    def test___call___with_is_stream(self, sleep):
```
parthea, line 721: Add a docstring to explain what we're checking for in this test.

google/api_core/retry_streaming.py
```diff
@@ -0,0 +1,183 @@
...
+            except Exception as exc:
+                self._handle_exception(exc)
+            # if exception was retryable, use new target for return value
+            return self.__next__()
+        else:
+            raise AttributeError(
+                "send() not implemented for {}".format(self.active_target)
+            )
+
+    def throw(self, typ, val=None, tb=None):
```
parthea, line 161: Please add type hints for these args. Please document the args in the docstring. Use more descriptive arg names. eg. What is `tb`?

google/api_core/retry_streaming.py
```diff
@@ -0,0 +1,183 @@
...
+          - AttributeError if the active_target does not have a close() method
+        """
+        if getattr(self.active_target, "close", None):
+            return self.active_target.close()
+        else:
+            raise AttributeError(
+                "close() not implemented for {}".format(self.active_target)
+            )
+
+    def send(self, value):
```
parthea, line 136: Add type hint for return value. Add information in the docstring to describe the arg `value`.

google/api_core/retry_streaming_async.py
```diff
@@ -0,0 +1,234 @@
...
+                    self.remaining_timeout_budget -= next_sleep
+            _LOGGER.debug(
+                "Retrying due to {}, sleeping {:.1f}s ...".format(exc, next_sleep)
+            )
+            # sleep before retrying
+            await asyncio.sleep(next_sleep)
+            self.active_target = None
+            await self._ensure_active_target()
+
+    def _subtract_time_from_budget(self, start_timestamp):
```
parthea, line 117: Add type hints for `start_timestamp`

google/api_core/retry_streaming_async.py
```diff
@@ -0,0 +1,234 @@
...
+        Args:
+        - start_timestamp (datetime): The time at which the last operation
+            started.
+        """
+        if self.remaining_timeout_budget is not None:
+            self.remaining_timeout_budget -= (
+                datetime_helpers.utcnow() - start_timestamp
+            ).total_seconds()
+
+    async def _iteration_helper(self, iteration_routine: Awaitable):
```
parthea, line 131: Add type hints for return value. Document return value in docstring.

parthea (commented):
Added minor observations.  Requesting review from either @vchudnov-g or @atulep

google/api_core/retry_async.py
```diff
@@ -224,11 +235,29 @@ async def retry_wrapped_func(*args, **kwargs):
                 target,
                 self._predicate,
                 sleep_generator,
-                self._timeout,
+                timeout=self._timeout,
+                on_error=on_error,
+            )
+
+        @functools.wraps(func)
+        def retry_wrapped_stream(*args, deadline_dt=None, **kwargs):
```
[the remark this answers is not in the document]
daniel-sanche: Hmm I don't believe `deadline_dt` is needed anymore, I'll remove it

google/api_core/retry_streaming.py
daniel-sanche: `send`, `throw`, `next`, and `close` are all part of the generator spec, so this signature comes from there. 

I'm a bit hesitant to spend too much time documenting the generator spec here. Maybe we should just use *args and **kwargs and pass them instead?

google/api_core/retry_streaming_async.py
```diff
@@ -0,0 +1,266 @@
...
+                "Timeout of {:.1f}s exceeded".format(self.timeout),
+                None,
+            )
+        try:
+            # start the timer for the current operation
+            start_timestamp = datetime_helpers.utcnow()
+            # grab the next value from the active_target
+            next_val_routine = asyncio.wait_for(
+                iteration_routine, self.remaining_timeout_budget
+            )
```
daniel-sanche, line 175: After doing some bench marking, it looks like `wait_for` slows things down significantly.

I'm thinking it might be best to have this follow the approach used by the synchronous version, and only timeout after the await is finished, rather than interrupting it early

Alternatively, Python3.11 has new timeout features which I believe are more performant, (along with a [third-party library](https://github.com/aio-libs/async-timeout) that is backwards compatible). But probably best to keep things simple.

Let me know what you think

google/api_core/retry.py
```diff
@@ -154,7 +155,7 @@ def retry_target(
     higher-level retry helper :class:`Retry`.
 
     Args:
-        target(Callable): The function to call and retry. This must be a
+        target(Callable[None, Any]): The function to call and retry. This must be a
```
atulep, line 158: What do these correspond do ? It's a little weird to have None as an arg

google/api_core/retry.py
```diff
@@ -301,6 +302,15 @@ class Retry(object):
...
         multiplier (float): The multiplier applied to the delay.
         timeout (float): How long to keep retrying, in seconds.
+        on_error (Callable[Exception]): A function to call while processing
+            a retryable exception. Any error raised by this function will
+            *not* be caught.
+        is_stream (bool): Indicates whether the input function
+            should be treated as an stream function (i.e. a Generator,
+            or function that returns an Iterable). If True, the iterable
+            will be wrapped with retry logic, and any failed outputs will
+            restart the stream. If False, only the input function call itself
```
atulep, line 312: Restart the stream from the last non-failed output, right ?

dizcology: Since this functionality is not intended for external users directly, but rather for the library generator and handwritten library code, this seems fine.  But the generator and library developers need to be aware of this and design their `target` callable carefully if they want to accomplish certain behaviors such as the "filter through modified request" approach you described.

Please expand the docstring of `RetryableGenerator` to provide some guidance and example pattern for the library developers so they do not make false assumptions.

google/api_core/retry_streaming_async.py
```diff
@@ -0,0 +1,260 @@
...
+        """
+        # check for expired timeouts before attempting to iterate
+        self._check_timeout(datetime_helpers.utcnow())
+        try:
+            # grab the next value from the active_target
+            # Note: interrupting with asyncio.wait_for is expensive,
+            # so we only check  for timeouts at the start of each iteration
+            return await iteration_routine
+        except (Exception, asyncio.CancelledError) as exc:
+            await self._handle_exception(exc)
```
dizcology, line 177: Do we want to do anything special in `_handle_exception` if the exception is `CancelledError`?

dizcology (approved):

google/api_core/retry_streaming_async.py
daniel-sanche: Good question... The default behavior is that cancellations will kill a generator. I added the extra CancelledError catch so that [cancellations could be retryable](https://github.com/googleapis/python-api-core/blob/14c195cac19606870e0d43c71a2c5409b7ba5811/tests/asyncio/test_retry_async.py#L546). (They usually aren't eligible for retries since they inherit from BaseException instead of Exception)

Looking at it again though, I'm not sure if this belongs in the wrapper. Users can always add a try/catch inside their wrapped generators if they want special handling for cancellations, and that probably makes more sense than handling it here. I think I'll remove this extra catch

google/api_core/retry.py
daniel-sanche: Sounds good, I added a much more documentation. Let me know that looks better

So far this will mostly just be useful for hand-written layers. If we want to add this to gapic libraries automatically, we could try to classify the different kinds of streams our libraries return and provide a couple different presets for different kinds of streams. But so far that's been out of scope.

parthea (approved):

google/api_core/retry_async.py
```diff
@@ -224,11 +236,29 @@ async def retry_wrapped_func(*args, **kwargs):
...
+            return retry_target_stream(
+                target,
+                self._predicate,
+                sleep_generator,
+                timeout=self._timeout,
                 on_error=on_error,
             )
 
-        return retry_wrapped_func
+        if self._is_stream:
```
vchudnov-g, line 258: we can't use a function pointer here, like you did in the sync case (`retry_func = retry_target_stream if self._is_stream else retry_target`), so that we only need one wrapped function instead of two that are almost identical?

NB: I notice that you're making changing to this PR, so I can't find the code that I quoted in the current version, though it is in the PR that I cloned locally

vchudnov-g (changes requested):
I focused on the sync code and left some comments, but I noticed the PR was changing out from under me, so I'll continue reviewing once it quiets down for a bit. Hopefully my comments still make sense. ;-)

google/api_core/retry_streaming.py
```diff
@@ -0,0 +1,221 @@
...
+    ```
+    def attempt_with_modified_request(target, request, seen_items=[]):
+        # remove seen items from request on each attempt
+        new_request = modify_request(request, seen_items)
+        new_generator = target(new_request)
+        for item in new_generator:
+            yield item
+            seen_items.append(item)
+
+    retry_wrapped = Retry(is_stream=True)(attempt_with_modified_request, target, request, [])
```
vchudnov-g, line 51: Hmm, I'm not sure why this comment didn't make on my previous review:

`Retry` takes a unary target, so I would tweak the example as follows. WDYT?

```py
def attempt_with_modified_request(target, request):
    seen_items=[]
    def attempter(target):
        # remove seen items from request on each attempt
        new_request = modify_request(request, seen_items)
        new_generator = target(new_request)
        for item in new_generator:
            yield item
            seen_items.append(item)
    return attempter

retry_wrapped = Retry(is_stream=True)(attempt_with_modified_request(target, request))

```

google/api_core/retry.py
```diff
@@ -314,6 +315,18 @@ class Retry(object):
         maximum (float): The maximum amount of time to delay in seconds.
         multiplier (float): The multiplier applied to the delay.
         timeout (float): How long to keep retrying, in seconds.
+        on_error (Callable[Exception]): A function to call while processing
```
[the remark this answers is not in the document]
daniel-sanche: Yeah I also thought that was surprising. I added a comment

google/api_core/retry_streaming.py
```diff
@@ -0,0 +1,220 @@
...
+    if is_timeout:
+        return (
+            exceptions.RetryError(
+                "Timeout of {:.1f}s exceeded".format(timeout_val),
+                src_exc,
+            ),
+            src_exc,
+        )
+    else:
+        return exc_list[-1], None
```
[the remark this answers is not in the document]
daniel-sanche: I think the problem was with mypy, since src_exc can be None but we want to always return an exception.

I made some changes to fall back to a default error if no exceptions are passed though. Let me know what you think

google/api_core/retry_streaming.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+
+    for sleep in sleep_generator:
+        # Start a new retry loop
+        try:
+            # create and yeild from a new instance of the generator from input generator function
+            subgenerator = target()
+            return (yield from subgenerator)
+        # handle exceptions raised by the subgenerator
+        except Exception as exc:
```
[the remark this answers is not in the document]
daniel-sanche: For some reason I wasn't getting any lint issues here, but added

google/api_core/retry_async.py
daniel-sanche: I think the only reason is because the existing `retry_wrapped_func` was an async function, and I was trying to add functionality without changing too much of the existing code.

But I just made a change to make `retry_wrapped_func` into a sync function that returns the retry coroutine directly instead of awaiting it, and that makes things cleaner. So let me know if that  works

google/api_core/retry_streaming.py
daniel-sanche: Good catch. That change looks fine to me, but I might have a slight preference to just breaking the end into two lines to avoid the extra nesting:

```python
    def attempt_with_modified_request(target, request, seen_items=[]):
        # remove seen items from request on each attempt
        new_request = modify_request(request, seen_items)
        new_generator = target(new_request)
        for item in new_generator:
            yield item
            seen_items.append(item)

    retry_wrapped_fn = Retry(is_stream=True)(attempt_with_modified_request)
    retryable_generator = retry_wrapped_fn(target, request)
```

Let me know what you think

google/api_core/retry_streaming.py
```diff
@@ -0,0 +1,221 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Generator wrapper for retryable streaming RPCs.
+This function will be used when initilizing a retry with
+``Retry(is_stream=True)``.
+
+When ``is_stream=False``, the target is treated as a callable,
```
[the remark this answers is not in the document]
daniel-sanche: Yeah, my only worry is it would be adding a lot of text to the existing wall of text in `Retry`. I thought it would be better to give a brief description and say "for more information, see retry_target_stream" instead.

Do you think this should be moved into the main Retry docstring?

google/api_core/retry_streaming.py
```diff
@@ -0,0 +1,220 @@
...
+            error_list.append(exc)
+            if not predicate(exc):
+                final_exc, source_exc = exc_factory(
+                    exc_list=error_list, is_timeout=False
+                )
+                raise final_exc from source_exc
+            if on_error is not None:
+                on_error(exc)
+        finally:
+            if subgenerator is not None and getattr(subgenerator, "close", None):
```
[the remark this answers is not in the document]
daniel-sanche: Good catch. I believe this is left-over from an earlier attempt, where I was trying to implement more of the generator protocol manually instead of relying on `yield from` (I  originally wanted to implement timeouts in between yields as well)

I think it should be safe to remove this extra check, and the unit tests seem to back that up

google/api_core/retry_streaming.py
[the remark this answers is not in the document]
vchudnov-g: Yes, I think that's the right level of abstraction for these comments. I suggest keeping the more immediate, more likely-to-be-referenced parts at the top of the comment, and these details further down

google/api_core/retry_streaming.py
vchudnov-g: OK, I think that should work. Had to spend some time on the various levels of indirection, with the decorators and generators.

google/api_core/retry_async.py
vchudnov-g: I like the new structure

google/api_core/retry_streaming.py
daniel-sanche: Yeah I'm fine with either way, whatever seems more readable to you

parthea:
Waiting on showcase test https://github.com/googleapis/gapic-generator-python/pull/1764

google/api_core/retry_streaming.py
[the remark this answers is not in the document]
daniel-sanche: done

daniel-sanche:
> Waiting on showcase test [googleapis/gapic-generator-python#1764](https://github.com/googleapis/gapic-generator-python/pull/1764)

Actually I think that PR is blocked on this one. Most tests won't pass there until this feature is merged

tests/unit/test_retry.py
```diff
@@ -467,3 +490,425 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+            on_error=on_error,
+            predicate=retry.if_exception_type(ValueError),
+            is_stream=True,
+            timeout=None,
+        )
+        result = retry_(self._generator_mock)(error_on=3)
+        # error thrown on 3
+        # generator should contain 0, 1, 2 looping
+        unpacked = [next(result) for i in range(10)]
+        assert unpacked == [0, 1, 2, 0, 1, 2, 0, 1, 2, 0]
```
vchudnov-g, line 564: Where does `_generator_mock.i` get reset when you `raise ValueError(…)`?

vchudnov-g (commented):
I _think_ the sync logic is fine. In the middle of reviewing the tests, but left some comments and questions. I'll resume my review from the sync tests  and then go on to the async code.

google/api_core/retry.py
```diff
@@ -314,6 +426,18 @@ class Retry(object):
...
```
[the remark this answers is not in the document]
daniel-sanche: Fixed. This docstring was taken from a previous place in the file and should be unrelated to the changes in this PR but, I updated it everywhere

tests/unit/test_retry.py
```diff
@@ -467,3 +490,425 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
         target.assert_has_calls([mock.call("meep"), mock.call("meep")])
         sleep.assert_any_call(retry_._initial)
+
+    def _generator_mock(
+        self,
+        num=5,
+        error_on=None,
+        return_val=None,
+        exceptions_seen=None,
+        ignore_sent=False,
```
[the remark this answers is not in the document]
daniel-sanche: It's basically just needed for one test, where I want to make sure `generator.send()` works properly. I simplified things by just using a different mock generator in that test

tests/unit/test_retry.py
```diff
@@ -467,3 +490,425 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        """
+        Helper to create a mock generator that yields a number of values
+        Generator can optionally raise an exception on a specific iteration
+        """
+        try:
+            sent_in = None
+            for i in range(num):
+                if error_on and i == error_on:
+                    raise ValueError("generator mock error")
+                sent_in = yield (sent_in if sent_in else i)
```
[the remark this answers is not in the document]
daniel-sanche: The mock generator would return an incrementing i unless both something was sent in on the last yield, and ignore_sent == False

But yeah, this is needlessly complicated to support one or two tests that use it, so I removed the whole `sent_in` functionality

tests/unit/test_retry.py
daniel-sanche: This is testing the core functionality of the streaming retries implementation. When the target generator raises an exception, a new instance is [swapped out under the hood](https://github.com/googleapis/python-api-core/blob/8cc6ea9c4f9717a2a0b0c8e2bd55efc726f5837e/google/api_core/retry_streaming.py#L104), without disrupting the wrapper stream

tests/unit/test_retry.py
```diff
@@ -467,3 +490,425 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+                if ignore_sent:
+                    sent_in = None
+            return return_val
+        except (Exception, BaseException, GeneratorExit) as e:
+            # keep track of exceptions seen by generator
+            if exceptions_seen is not None:
```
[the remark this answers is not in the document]
daniel-sanche: Actually I had to revert this; we want to modify the list that was passed in instead of replacing the list here.

I added some documentation around this function

vchudnov-g:
I will continue reviewing this shortly, but we also want to get https://github.com/googleapis/gapic-generator-python/pull/1764 ready to submit immediately thereafter so that these changes are covered by tests. That, in turn, requires addressing https://github.com/googleapis/gapic-showcase/issues/1377, which in turn requires at least addressing https://github.com/googleapis/gapic-showcase/issues/1378 (assigned to me)

daniel-sanche:
> I will continue reviewing this shortly, but we also want to get [googleapis/gapic-generator-python#1764](https://github.com/googleapis/gapic-generator-python/pull/1764) ready to submit immediately thereafter so that these changes are covered by tests. That, in turn, requires addressing [googleapis/gapic-showcase#1377](https://github.com/googleapis/gapic-showcase/issues/1377), which in turn requires at least addressing [googleapis/gapic-showcase#1378](https://github.com/googleapis/gapic-showcase/issues/1378) (assigned to me)

We could skip the problematic tests in https://github.com/googleapis/gapic-generator-python/pull/1764 for now if needed, to unblock this one. Let me know

dizcology:
@vchudnov-g: your comment https://github.com/googleapis/python-api-core/pull/495#issuecomment-1767039543 clearly explains a chain of blocking issues.

Since the features do not encounter the same issues with the actual service, can we move the feature PRs (https://github.com/googleapis/python-api-core/pull/495 and https://github.com/googleapis/gapic-generator-python/pull/1764) forward while working on the showcase issue https://github.com/googleapis/gapic-showcase/issues/1378 separately, in parallel?

tests/asyncio/test_retry_async.py
```diff
@@ -402,3 +411,492 @@ async def test___init___when_retry_is_executed(self, sleep, uniform):
...
+            return CustomIterable()
+
+        if awaitale_wrapped:
+
+            async def wrapper():
+                return iterable_fn()
+
+            decorated = retry_(wrapper)
+        else:
+            decorated = retry_(iterable_fn)
```
vchudnov-g, line 712:
```suggestion
        if awaitable_wrapped:
            async def wrapper():
                return iterable_fn()
            decorated = retry_(wrapper)
        else:
            decorated = retry_(iterable_fn)
```
```suggestion
        if awaitale_wrapped:

            async def wrapper():
                return iterable_fn()

            decorated = retry_(wrapper)
        else:
            decorated = retry_(iterable_fn)
```

tests/asyncio/test_retry_async.py
```diff
@@ -402,3 +411,492 @@ async def test___init___when_retry_is_executed(self, sleep, uniform):
...
+
+        retryable = await decorated()
+        result = await retryable.__anext__()
+        assert result == 0
+        await retryable.asend("test") == 1
+        await retryable.asend("test2") == 2
+        await retryable.asend("test3") == 3
+
+    @pytest.mark.parametrize("awaitale_wrapped", [True, False])
```
vchudnov-g, line 721:
```suggestion
    @pytest.mark.parametrize("awaitable_wrapped", [True, False])
```

tests/asyncio/test_retry_async.py
```diff
@@ -402,3 +411,492 @@ async def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        await retryable.aclose()
+        with pytest.raises(StopAsyncIteration):
+            await retryable.__anext__()
+        # try closing new generator
+        new_retryable = await decorated()
+        await new_retryable.aclose()
+        with pytest.raises(StopAsyncIteration):
+            await new_retryable.__anext__()
+
+    @pytest.mark.parametrize("awaitale_wrapped", [True, False])
```
vchudnov-g, line 765:
```suggestion
    @pytest.mark.parametrize("awaitable_wrapped", [True, False])
```

vchudnov-g (commented):
Reviewed the the async logic. I think it looks god, but I have some questions, partly due to learning about Python async stuff.

google/api_core/retry_streaming_async.py
```diff
@@ -0,0 +1,188 @@
...
+            try:
+                gen_instance = await gen_instance  # type: ignore
+            except TypeError:
+                # was not awaitable
+                pass
+            subgenerator = cast(AsyncIterable["_Y"], gen_instance).__aiter__()
+
+            # if target is a generator, we will advance it using asend
+            # otherwise, we will use anext
+            supports_send = bool(getattr(subgenerator, "asend", None))
```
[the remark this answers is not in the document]
daniel-sanche: There's some value in letting someone build a custom object that supports `aclose`, but doesn't care about the others. But yeah, it's probably not worth the extra complexity/perf cost. I made the change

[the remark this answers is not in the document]
daniel-sanche: Good point, fixed

google/api_core/retry_streaming_async.py
```diff
@@ -0,0 +1,188 @@
...
+                    next_value = await subgenerator.asend(sent_in)  # type: ignore
+                else:
+                    next_value = await subgenerator.__anext__()
+                ## Yield from Wrapper to caller
+                try:
+                    # yield last value from subgenerator
+                    # exceptions from `athrow` and `aclose` are injected here
+                    sent_in = yield next_value
+                except GeneratorExit:
+                    # if wrapper received `aclose`, pass to subgenerator and close
```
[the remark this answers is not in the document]
daniel-sanche: I actually mean the outermost object here, not the subgenerator. We need to pass it down to the subgenerator

Let me know if there's a better word

```
async def inner_fn():
  yield 0
  yield 1
wrapped_fn = retry_target(inner_fn)
await wrapped_fn.__anext__()
await wrapped_fn.aclose()
```

tests/asyncio/test_retry_async.py
```diff
@@ -402,3 +411,492 @@ async def test___init___when_retry_is_executed(self, sleep, uniform):
...
+    async def test___call___generator_cancellations(self):
+        """
+        cancel calls should propagate to the generator
+        """
+        # test without cancel as retryable
+        retry_ = retry_async.AsyncRetry(is_stream=True)
+        utcnow = datetime.datetime.utcnow()
+        mock.patch("google.api_core.datetime_helpers.utcnow", return_value=utcnow)
+        generator = await retry_(self._generator_mock)(sleep_time=0.2)
+        await generator.__anext__() == 0
```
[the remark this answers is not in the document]
daniel-sanche: You can read this as (await generator.__anext__()) == 0. It's pulling the next value out of the generator, and comparing it to 0

tests/asyncio/test_retry_async.py
```diff
@@ -402,3 +411,492 @@ async def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        task = asyncio.create_task(generator.__anext__())
+        await asyncio.sleep(0.1)
```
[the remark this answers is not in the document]
daniel-sanche: Hmm I think I thought it needed a chance context switch to the task, but actually that shouldn't be needed here. Good catch, removed

tests/asyncio/test_retry_async.py
```diff
@@ -402,3 +411,492 @@ async def test___init___when_retry_is_executed(self, sleep, uniform):
...
+            predicate=retry_async.if_exception_type(ValueError),
+            is_stream=True,
+            timeout=None,
+        )
+        generator = await retry_(self._generator_mock)(error_on=3)
+        with pytest.raises(TypeError) as exc_info:
+            await generator.asend("can not send to fresh generator")
+            assert exc_info.match("can't send non-None value")
+
+        # error thrown on 3
```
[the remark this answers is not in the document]
daniel-sanche: The other one was doing next: `[await generator.__anext__() for i in range(10)]`, this one is using asend: `[await generator.asend(i) for i in range(10)]`

Or are you thinking of something else?

tests/asyncio/test_retry_async.py
```diff
@@ -402,3 +411,492 @@ async def test___init___when_retry_is_executed(self, sleep, uniform):
...
+            await generator.__anext__()
+        # should retry if throw retryable exception
+        exception_list = []
+        generator = await decorated(10, exceptions_seen=exception_list)
+        for i in range(2):
+            await generator.__anext__()
+        throw_val = await generator.athrow(ValueError("test"))
+        assert throw_val == 0
+        assert isinstance(exception_list[0], ValueError)
+        # calling next on closed generator should not raise error
```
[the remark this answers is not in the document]
daniel-sanche: correct

tests/unit/test_retry.py
```diff
@@ -487,3 +510,434 @@ def test___init___when_retry_is_executed(self, sleep, uniform):
...
+        with pytest.raises(ValueError):
+            # generator should not retry if error is thrown on yield
+            gen = auto_retry_(wrapped)()
+            [next(gen) for i in range(10)]
+        # force non-detected to be accepted as generator
+        gen = gen_retry_(wrapped)()
+        unpacked = [next(gen) for i in range(10)]
+        assert unpacked == [0, 1, 2, 3, 4, 5, 0, 1, 2, 3]
+
+    def test_exc_factory_non_retryable_error(self):
```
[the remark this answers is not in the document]
daniel-sanche: Possibly, although I was actually hoping to add the exception factory functionality to the other retry code as well in a subsequent PR. So maybe these tests could eventually target `retry.Retry` like the others. 

But I'm also not sure if it makes sense to expose in `retry.Retry`, or keep it lower level for library developers in `retry_target` and `retry_target_stream`.

Thoughts?

google/api_core/retry_async.py
```diff
@@ -196,15 +292,25 @@ def __init__(
...
+        self._is_stream = is_stream
 
-    def __call__(self, func, on_error=None):
+    def __call__(
+        self,
+        func: Callable[
+            ..., Awaitable[_R] | AsyncIterable[_Y] | Awaitable[AsyncIterable[_Y]]
+        ],
+        on_error: Callable[[BaseException], Any] | None = None,
+    ) -> Callable[_P, Awaitable[_R | AsyncGenerator[_Y, None]]]:
```
daniel-sanche, line 303: I'm starting to wonder whether the streaming retries should use an entirely different class:
1) this type definition is complicated, and hard to deal with in downstream code
2) We treat Retry objects as something that end-users can build on their own, then pass into gapic methods. It doesn't really make sense to pass a retry object with `is_stream=True` to a unary method (or the other way around really, even though that's the default behaviour now)

Let me know what you think

daniel-sanche: I made the change, explained in a separate comment

daniel-sanche:
Reviewers: I made a couple changes recently:

### using separate classes for StreamingRetry and AsyncStreamingRetry

Previously, I tried to share the same retry objects with a flag, but this made typing very difficult, and didn't make a lot of sense with how the Retry objects are used. The Retry objects can be passed in to gapic functions to detail the retry policy, but it doesn't make sense to pass a Retry with `is_stream=True` to a unary function.

Now, there are 4 separate retry classes instead of 2: `Retry`, `AsyncRetry`, `StreamingRetry` and `AsyncStreamingRetry`. I created a new base class to contain the shared logic

[Commit](https://github.com/googleapis/python-api-core/pull/495/commits/19a998d88b3bbfc9a7cdf9b35c572925b19e0000)

### drop partial before retry_target_stream

Previously, we would use functools.partial to create a nullary function before passing it in to the retry wrapper. That makes sense for unary calls, but we will need more flexibility for streams. 

I imagine at some point we will want to be able to configure different retry policies, where the request arguments are modified for each attempt. This change keeps that option available without needing a breaking change in the future

[Commit](https://github.com/googleapis/python-api-core/pull/495/commits/e2d9c9cc2526f0305d137a5858658a2e26bbda80)

tests/unit/test_retry.py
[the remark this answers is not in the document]
daniel-sanche: I did [some refactoring](https://github.com/googleapis/python-api-core/pull/495#issuecomment-1821401234), which included moving these tests into a separate file

google/api_core/retry_async.py
```diff
@@ -74,23 +94,23 @@ async def check_if_exists():
 async def retry_target(
     target, predicate, sleep_generator, timeout=None, on_error=None, **kwargs
 ):
-    """Call a function and retry if it fails.
+    """Await a coroutine and retry if it fails.
```
vchudnov-g, line 97: IIUC, the sleep-delay logic should be the same between all four versions of this function: syncvs async, unary vs streaming. I would suggest:

* at the very least making the four functions look as similar as possible with their timeout handling logic. For example, the async versions uses `min()`, which the sync version probably should as well. The streaming functions use `exception_factory`; should the unary ones as well, for consistenct?

* consider whether it's possible and desirable to factor any of the time- and exception-handling functionality into a common function. I suspect the answer is "no" because it would make the code more complex and harder to follow, but it would have the advantage that we'd be using the same logic in the parts where it makes sense. Maybe just factor out the `if deadline` and `_LOGGER` at the bottom of the loop into a helper function that all four versions call to increment the sleep consistently?

google/api_core/retry_async.py
```diff
@@ -74,23 +94,23 @@ async def check_if_exists():
 async def retry_target(
     target, predicate, sleep_generator, timeout=None, on_error=None, **kwargs
```
vchudnov-g, line 95: the streaming versions have type declarations. Do you want to have type declarations in the non-stream versions?

vchudnov-g (commented):
Thanks for the refactoring! I think it makes the code and review a lot easier.

I think we're almost there. I left one substantive comment about ensuring the sleep increments are implemented consistently in all four versions, The other comments are more minor. 

I'll take a final look at the tests once we're satisfied with the main code.

(BTW: I suggest in the future only merging `main` before the initial review request and after approval is granted, so that it's easier for reviewers to see the changes during the review process without extraneous diffs coming in from the `main` merge. Then if that final `main` merge affected the PR's files, we can do quick re-review.)

google/api_core/retry.py
```diff
@@ -89,9 +94,55 @@ def check_if_exists():
 _ASYNC_RETRY_WARNING = "Using the synchronous google.api_core.retry.Retry with asynchronous calls may lead to unexpected results. Please use google.api_core.retry_async.AsyncRetry instead."
 
 
+class RetryFailureReason(Enum):
```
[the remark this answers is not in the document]
daniel-sanche: Ok, I did some reorganization, moving the files into their own `retry` folder. The TYPE_CHECKING was a bit more complicated, since import * doesn't play nicely with our linting (and the generics don't exist in the retry_base) , but I think it still lets us simplify things a bit

I also made sure to keep existing exported names the same, to leave existing code unchanged.

Let me know what you think

google/api_core/retry_async.py
daniel-sanche: I was trying to avoid making changes to the existing retries, but I suppose with the new base classes, it makes sense to generalize them a bit. I added a shared `_retry_error_helper` function, which either raises an error, or logs the retry message.

Note though that I had to make a (potentially breaking) change to the async unary retry logic to make it consistent though. Previously async_unary would attempt to cancel early when the deadline is reached, while the other retries only check the deadline after the request terminates. The old behavior was very slow (using asyncio.wait_for), and caused [race conditions](https://github.com/googleapis/python-api-core/issues/528). I'd consier it a bug, but fixing the behavior may count as a breaking change (even though gapic functions shouldn't be affected)

daniel-sanche:
@vchudnov-g I addressed the changes you suggested, and also removed the inspect.isawaitable check in retry_unary. It is an expensive check, and I think it's hard to justify having it run on every single rpc. But maybe we can add a flag or something for it?

tests/asyncio/retry/test_retry_streaming_async.py
```diff
@@ -0,0 +1,541 @@
...
+
+            async def wrapper():
+                return iterable_fn()
+
+            decorated = retry_(wrapper)
+        else:
+            decorated = retry_(iterable_fn)
+
+        retryable = await decorated()
+        result = await retryable.__anext__()
```
vchudnov-g, line 354: should this be `asend`? (Honest question; I'm still murky on this)

tests/asyncio/retry/test_retry_streaming_async.py
```diff
@@ -0,0 +1,541 @@
...
+        assert result == 0
+        await retryable.asend("test") == 1
```
vchudnov-g, line 356: Should these have `asserts`?

vchudnov-g (approved):
The refactoring was fantastic and made the code a lot cleaner. Thank you for that!

I have just a couple of minor code questions on the tests. The rest are doc comments and nits, so this PR LGTM.

IIUC, this changes the current (bad) behavior of the async retries, so this should be consdeired a breaking change? WDYT?

tests/asyncio/retry/test_retry_streaming_async.py
daniel-sanche: You can't call send until it's been "started" (i.e., next has been called once). I'll add a comment around this

google/api_core/retry/retry_base.py
```diff
@@ -0,0 +1,346 @@
...
+# considered a function.
+if_transient_error = if_exception_type(
+    exceptions.InternalServerError,
+    exceptions.TooManyRequests,
+    exceptions.ServiceUnavailable,
+    requests.exceptions.ConnectionError,
+    requests.exceptions.ChunkedEncodingError,
+    auth_exceptions.TransportError,
+)
+"""A predicate that checks if an exception is a transient API error.
```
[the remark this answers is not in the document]
daniel-sanche: This is not new code, I just moved the [existing function](https://github.com/googleapis/python-api-core/blob/46e878976a32d75a57e4d7995ed89629b6ef47bd/google/api_core/retry.py#L124) to this new shared file

If we do changes here, I'd probably prefer to open an issue and address it separately, since it's not related to streaming retries. Let me know what you think

google/api_core/retry/retry_base.py
```diff
@@ -0,0 +1,346 @@
...
+    - Logs the error
+
+    Args:
+       - exc: the exception that was raised
+       - deadline: the deadline for the retry, calculated as a diff from time.monotonic()
+       - next_sleep: the calculated next sleep interval
+       - error_list: the list of exceptions that have been raised so far
+       - predicate_fn: the predicate that was used to determine if the exception should be retried
+       - on_error_fn: the callback that was called when the exception was raised
+       - exc_factory_fn: the callback that was called to build the exception to be raised on terminal failure
```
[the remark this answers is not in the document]
daniel-sanche: I'm not sure about the `(a return value of `None` signals not to retry)` part. The type annotation just says to return `bool`. Technically it should work with any value based on its truthiness, but I'm not sure if we want to commit to supporting extra types in the docstrings

google/api_core/retry/retry_base.py
```diff
@@ -0,0 +1,346 @@
...
+    """
+    Base class for retry configuration objects. This class is intended to capture retry
+    and backoff configuration that is common to both synchronous and asynchronous retries,
+    for both unary and streaming RPCs. It is not intended to be instantiated directly,
+    but rather to be subclassed by the various retry configuration classes.
+    """
+
+    def __init__(
+        self,
+        predicate: Callable[[Exception], bool] = if_transient_error,
```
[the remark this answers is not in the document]
daniel-sanche: Yeah if it wasn't existing code I think it would be worth using a different name for it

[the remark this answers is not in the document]
daniel-sanche: `if_transient_error` is already exported publicly and may be used downstream though. 

I guess if we make this a breaking change, we can change the naming now. But IMO probably not worth it

google/api_core/retry/retry_base.py
```diff
@@ -0,0 +1,346 @@
...
+            be greater than 0.
+        maximum (float): The maximum amount of time to delay.
+        multiplier (float): The multiplier applied to the delay.
+
+    Yields:
+        float: successive sleep intervals.
+    """
+    delay = min(initial, maximum)
+    while True:
+        yield random.uniform(0.0, delay)
```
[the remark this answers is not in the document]
daniel-sanche: Yeah this code pre-dates the PR, and it seems like it was written to align with [this retryPolicy spec](https://github.com/grpc/proposal/blob/d4fc009e55f95297374e821d67d679b931753a59/A6-client-retries.md#integration-with-service-config)

google/api_core/retry/retry_unary.py
```diff
@@ -195,12 +134,12 @@ def retry_target(
...
-    if timeout is not None:
-        deadline = datetime_helpers.utcnow() + datetime.timedelta(seconds=timeout)
-    else:
-        deadline = None
-
-    last_exc = None
+    deadline = time.monotonic() + timeout if timeout is not None else None
+    error_list: list[Exception] = []
+    # make a partial with timeout applied
+    exc_factory = lambda e, t: (exception_factory or _build_retry_error)(  # noqa: E731
```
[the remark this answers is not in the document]
daniel-sanche: Possibly, but the main purpose of the lambda here is to lock in the timeout parameter. This makes it a bit easier/cleaner to build exceptions later, so we don't have to keep `timeout` in scope everywhere we raise exceptions

I guess this is less valuable now that all exceptions are handled in one place with `_retry_error_helper`. But it does cut down the list of arguments that are passed in to `_retry_error_helper` by by one

I'll take a look at this again

google/api_core/retry/retry_unary.py
```diff
@@ -320,40 +247,27 @@ class Retry(object):
         maximum (float): The maximum amount of time to delay in seconds.
         multiplier (float): The multiplier applied to the delay.
         timeout (float): How long to keep retrying, in seconds.
+        on_error (Callable[Exception]): A function to call while processing
```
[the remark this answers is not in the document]
daniel-sanche: Is this something that can be opened as a future issue, since it's unrelated to streaming retries?

google/api_core/retry/retry_unary.py
```diff
@@ -212,34 +151,18 @@ def retry_target(
...
-        time.sleep(sleep)
+            # if exception not raised, sleep before next attempt
+            time.sleep(sleep)
 
     raise ValueError("Sleep generator stopped yielding sleep values.")
 
 
-class Retry(object):
-    """Exponential retry decorator.
+class Retry(_BaseRetry):
```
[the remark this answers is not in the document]
daniel-sanche: I made a couple other changes to the wording as well. It used to imply that the deadline was the "maximum time for the rpc", but it can't actually interrupt in-progress sync rpcs. So I describe it as a "window in which retries will be attempted" instead

google/api_core/retry/retry_streaming_async.py
```diff
@@ -0,0 +1,324 @@
...
+        AsyncGenerator: A retryable generator that wraps the target generator function.
+
+    Raises:
+        ValueError: If the sleep generator stops yielding values.
+        Exception: a custom exception specified by the exception_factory if provided.
+            If no exception_factory is provided:
+                google.api_core.RetryError: If the deadline is exceeded while retrying.
+                Exception: If the target raises an error that isn't retryable.
+    """
+    # create frozen partial from original call args
```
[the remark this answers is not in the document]
daniel-sanche: Hmm that may have gotten left over from a refactor. Good catch, removed

google/api_core/retry/retry_streaming.py
```diff
@@ -0,0 +1,330 @@
...
+            # if exception not raised, sleep before next attempt
+            time.sleep(sleep)
+
+    raise ValueError("Sleep generator stopped yielding sleep values.")
+
+
+class StreamingRetry(_BaseRetry):
+    """Exponential retry decorator for streaming synchronous RPCs.
+
+    This class returns a Generator when called, which wraps the target
```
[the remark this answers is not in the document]
daniel-sanche: If we decide to make this a breaking change, maybe we can remove/vastly simplify the whole discussion about deadline vs timeout? My impression was it was originally added to transition the `deadline` variable to `timeout`. If we have a new major version, we can probably just drop most references to deadline, since it has been deprecated for a while (Maybe just keep parsing it out of kwargs for compatibility, but not mention it in the signatures/docstrings?). Then we can just keep a simple explanation of what timeout specifically means

If not, keeping it in the `Retry` class sounds good to me

tests/unit/retry/test_retry_unary.py
```diff
@@ -145,30 +109,25 @@ async def test_retry_target_warning_for_retry(utcnow, sleep):
...
     target = mock.Mock(side_effect=exception)
     # Setup the timeline so that the first call takes 5 seconds but the second
     # call takes 6, which puts the retry over the deadline.
-    utcnow.side_effect = [
-        # The first call to utcnow establishes the start of the timeline.
-        datetime.datetime.min,
-        datetime.datetime.min + datetime.timedelta(seconds=5),
-        datetime.datetime.min + datetime.timedelta(seconds=11),
-    ]
+    monotonic.side_effect = [0, 5, 11]
```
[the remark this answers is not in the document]
daniel-sanche: Good point, but I also want to make sure we have some tests that use `deadline`, if that's supposed to be maintained as an alias. I added an extra `paramaterize` to check with both

But if we make this a breaking change, maybe we can drop deadline

tests/unit/retry/test_retry_unary.py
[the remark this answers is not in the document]
daniel-sanche: Hmm I just see the one test that uses `@pytest.mark.asyncio`, and it's the test that a warning is triggered when a coroutine is used with the sync Retry class, so it is necessary there. (Btw, [I think this functionality should be removed](https://github.com/googleapis/python-api-core/issues/542#issuecomment-1835269813), but that's a separate discussion)

Am I missing others?

tests/asyncio/retry/test_retry_unary_async.py
```diff
@@ -133,108 +130,17 @@ async def test_retry_target_bad_sleep_generator():
...
-            on_error=_some_function,
-        )
-        assert retry_._predicate == mock.sentinel.predicate
-        assert retry_._initial == 1
-        assert retry_._maximum == 2
-        assert retry_._multiplier == 3
-        assert retry_._deadline == 4
-        assert retry_._on_error is _some_function
+@pytest.mark.asyncio
+async def test_retry_streaming_target_bad_sleep_generator():
```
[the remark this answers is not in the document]
daniel-sanche: good catch, moved to streaming test file

tests/asyncio/retry/test_retry_streaming_async.py
daniel-sanche: Good catch, I believe they should be

google/api_core/retry/retry_base.py
```diff
@@ -0,0 +1,346 @@
...
+def _retry_error_helper(
+    exc: Exception,
+    deadline: float | None,
+    next_sleep: float,
+    error_list: list[Exception],
+    predicate_fn: Callable[[Exception], bool],
+    on_error_fn: Callable[[Exception], None] | None,
+    exc_factory_fn: Callable[
+        [list[Exception], RetryFailureReason],
+        tuple[Exception, Exception | None],
```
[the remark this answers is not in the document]
daniel-sanche: Yeah I think that makes more sense, changed

google/api_core/retry/retry_unary.py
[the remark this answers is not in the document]
daniel-sanche: Yeah, after moving the _build_retry_error to be a default arg, it looks cleaner to remove the whole lambda and add the extra arg to _retry_error_helper. I made that change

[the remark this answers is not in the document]
daniel-sanche: No longer needed with previous changes

daniel-sanche:
I addressed the latest round of changes

> IIUC, this changes the current (bad) behavior of the async retries, so this should be consdeired a breaking change? WDYT?

The the previous behaviour was 1) undocumented and not covered in the signature or docstrings, and 2) (I don't believe) widely used. So it's a bit of a judgement call. 

I'll mark this PR as breaking to be on the safer side, but I think there's a strong argument that breaking isn't needed if there are reasons to prefer a minor release. Let me know what you think

daniel-sanche:
> Can the breaking change be avoided by deprecating functionality instead of changing it?

Hmm yeah I guess it could, if we want to keep two versions of the AsyncRetry class hanging around for a while

As I mentioned above though, IMO this doesn't have to be a breaking change. The only breaking part is changing an undocumented behaviour, which would only [manifest as a bug](https://github.com/googleapis/python-api-core/issues/528) when used with our gapic libraries, which have their own timeout logic.

But I'll let you make that call. Let me know which option you prefer

google/api_core/retry/retry_base.py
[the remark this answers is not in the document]
vchudnov-g: I'm not seeing the change. (?)

google/api_core/retry/retry_base.py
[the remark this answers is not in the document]
vchudnov-g: Hmm, I see your point, but then we're not accurately describing how this function uses the preidcate. If a non-`True` truthy value is returned, the function will retry. How about 'predicate that takes `exc` and returns "true" if the operation should be retried'? Notice I lower-cased true to imply that it's actually truthiness (too subtle? lol), and I said "retry the operation", not "retry the exception".

google/api_core/retry/retry_base.py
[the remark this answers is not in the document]
vchudnov-g: Agreed.

google/api_core/retry/retry_unary.py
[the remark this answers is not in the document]
vchudnov-g: Sure.

google/api_core/retry/retry_unary.py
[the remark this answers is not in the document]
vchudnov-g: I like it.

google/api_core/retry/retry_streaming.py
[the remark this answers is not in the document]
vchudnov-g: That's a good point. We should decide about the breaking change issue.

tests/unit/retry/test_retry_unary.py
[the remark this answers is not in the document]
vchudnov-g: I like the parametrization idea! 

I think what you have is good. One could go a bit further but I think it would be less clear than what you have: 

```
@pytest.mark.parametrize("timeout_arg", ["deadline", "timeout"])
...

 kwargs = {timeout_arg: 10}
```

google/api_core/retry/retry_streaming.py
```diff
@@ -0,0 +1,263 @@
...
+    target: Callable[_P, Iterable[_Y]],
+    predicate: Callable[[Exception], bool],
+    sleep_generator: Iterable[float],
+    timeout: Optional[float] = None,
+    on_error: Optional[Callable[[Exception], None]] = None,
+    exception_factory: Callable[
+        [List[Exception], RetryFailureReason, Optional[float]],
+        Tuple[Exception, Optional[Exception]],
+    ] = _build_retry_error,
+    init_args: _P.args = (),
```
vchudnov-g, line 62: Why do the the sync and async versions of `retry_target_stream` have `init_args` and `init_kwargs` but the unary `retry_target`s do not? Shouldn't we be consistent?

vchudnov-g (commented):
OK, looking at this again to try to settle the breaking change determination. I have two additional code comments in the files.

On the breaking change determination:

1.  this PR makes streaming retries more reliable, so any code users previously had to manually enforce retries on streams will now be needed less or not at all, but it won't be broken by this change. Correct? 
2. The other change os that exceptions are propagated through the retry block, so users may now be seeing exceptions that they did not see before. Correct?

Item 1 by itself would suggest marking this non-breaking, though it may be good to call attention to the change so users can simplify parts of their code that may no longer be needed.
Item 2 seems like a breaking behavioral change that may surprise users.

I think we should flag this as a breaking change, and make it clear at the top of the PR description/merge-commit message that the only expected behavioral changes that may require action on the user's side are the new exceptions they may now receive from failing retries.

Thoughts? @daniel-sanche @parthea

google/api_core/retry/retry_streaming.py
daniel-sanche: I tried to explain that in the [comments here](https://github.com/googleapis/python-api-core/blob/85b3e02f30dc45b8fb2423f25eb0842380c38d10/google/api_core/retry/retry_streaming.py#L114)

Basically, every time you retry a unary rpc, it will use the same arguments, so it makes sense to lock them in to a partial before calling `retry_target`

With streaming rpcs, retries are more complicated, because we'd like to be able to recover the stream at the point it broke, instead of from the beginning. But the way to do that is different for each rpc. It's out of scope for this PR, but I can imagine at some point wanting to add a `ResumptionStrategy` class, and provide some kind of call-back that allows you to modify requests between attempts. For that, we need to keep the method and it's call args separate when starting `retry_target_stream`

We could change `retry_target` to work the same way to be consistent, but that would definitely be a breaking change. And it's unnecessary, because unary retries don't have the same problems as streams

TL;DR: Using a partial for `retry_target_stream` would paint us into a corner, so I'm trying to make things flexible for future improvements

google/api_core/retry/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+from .retry_streaming import StreamingRetry
+from .retry_streaming import retry_target_stream
+from .retry_streaming_async import AsyncStreamingRetry
+from .retry_streaming_async import retry_target_stream as retry_target_stream_async
+
+__all__ = (
+    "exponential_sleep_generator",
+    "if_exception_type",
+    "if_transient_error",
+    "_build_retry_error",
```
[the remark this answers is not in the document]
daniel-sanche: good point, removed

google/api_core/retry/retry_base.py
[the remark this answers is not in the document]
daniel-sanche: Opened here: https://github.com/googleapis/python-api-core/issues/569

google/api_core/retry/retry_base.py
```diff
@@ -0,0 +1,346 @@
...
+
+def exponential_sleep_generator(
+    initial: float, maximum: float, multiplier: float = _DEFAULT_DELAY_MULTIPLIER
+):
+    """Generates sleep intervals based on the exponential back-off algorithm.
+
+    This implements the `Truncated Exponential Back-off`_ algorithm.
+
+    .. _Truncated Exponential Back-off:
+        https://cloud.google.com/storage/docs/exponential-backoff
```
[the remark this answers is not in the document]
daniel-sanche: https://github.com/googleapis/python-api-core/issues/570

google/api_core/retry/retry_base.py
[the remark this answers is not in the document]
daniel-sanche: Yeah that works for me, I changed the text

daniel-sanche:
> 1. this PR makes streaming retries more reliable, so any code users previously had to manually enforce retries on streams will now be needed less or not at all, but it won't be broken by this change. Correct?
> 2. The other change os that exceptions are propagated through the retry block, so users may now be seeing exceptions that they did not see before. Correct?

I don't believe anything should change for users for either of these points. This PR is adding the new RetryStreaming and RetryStreamingAsync classes to `api_core`, but nothing will change downstream until we make changes in the downstream gapic_generator to use them.

gapic_generator currently [uses the unary retries.Retry](https://github.com/googleapis/python-logging/blob/2384981c9137a57a647a69a32b67dcacd619ea0a/google/cloud/logging_v2/services/logging_service_v2/transports/base.py#L216) object for all rpcs, and that won't change after this PR is merged. Once this functionality is in place, the next step would be looking into bringing it to gapic (but that will actually be a whole separate project on its own, because streaming rpcs will require different semi-custom resumption strategies. I'm not currently planning on taking this on myself, but it ties in with work Leah Cole has been doing in Node)

The main motivation of this change is making the retry streaming functionality available for veneer developers, until we figure out how to automate it for streaming rpcs in gapic libraries

---

The part that I think could be considered breaking is

3. AsyncRetry [used to use `asyncio.wait_for`](https://github.com/googleapis/python-api-core/blob/9866183b42f632dddade9ba6c69f9b3015118dd4/google/api_core/retry_async.py#L121) to cancel the attempt at the deadline, but that [caused race-condition issues](https://github.com/googleapis/python-api-core/issues/528), and wasn't consistent with how the other retries worked. I removed that as part of the refactoring we did to share logic between all retry classes

Our options are:

a. mark this as a breaking change
b. consider it a non-breaking change, since it is an 1) undocuemnt behaviour change to 2) fix a bug 3) that we don't have reason to believe would impact the workflow of many users
c. undo some of the recent refactoring to make async retries work the same as before, and address the race condition (and potential breaking change issues) in a different PR, so we can merge this one now

I'm fine with any option, just let me know which you both prefer

google/api_core/retry/retry_streaming.py
vchudnov-g: OK, makes sense. I wonder whether we should a brief note to that effect in the documentation for the two parameters ("these two parameters are provided in case the user needs to change the arguments in subsequent retry attempts"). But we can defer that for now.

vchudnov-g:
OK, I see. 
- On point 1, we agree.
- On point 2: for unaries, we were propagating the exceptions before, so no change there. And for streaming: these are new retry classes, so they can't break anyone
- On point 3: The behavior changes, but the exceptions that are raised out of the new retry code are exceptions like `RetryError` that the user should already be handling, and the current user code should be ready to handle either type of exception mentioned in the race condition in #528 . In other words, existing user code will not be broken, even if the particular exceptions that happen to be raised under a given set of circumstances would be different with or without the changes in this PR.

If this is an accurate summary, then I think calling this PR non-breaking is fine, which would result in this `feat` becoming a new minor release. @parthea WDYT? Would you like to do a pre-release for this minor release? I think we don't need to, but I'm totally open to being extra-cautious. FWIW, the existing tests, though refactored, all pass with this new code. 

Separately, I'll file an issue to have `gapic-generator-python` use the new retry streaming classes you implemented.

vchudnov-g:
One point of clarification: the current GAPICs do use `AsyncRetry` ([link](https://github.com/googleapis/gapic-generator-python/blob/f4e0389870a9e4725310ea8a0b01584dadaf7aec/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/async_client.py.j2#L24))

google/api_core/retry/retry_streaming_async.py
```diff
@@ -163,11 +163,15 @@ async def retry_target_stream(
 
                         # TODO: Remove this conditional once the minimum supported Python version is 3.11
                         if sys.version_info[:3] >= (3, 11, 0):
-                            await cast(AsyncGenerator["_Y", None], target_iterator).athrow(
-                                sys.exception())  # type: ignore
+                            await cast(
+                                AsyncGenerator["_Y", None], target_iterator
+                            ).athrow(
+                                sys.exception()  # type: ignore
+                            )
```
daniel-sanche, line 170: it looks like we could just use sys.exc_info()[1] to be compatible with all versions?

`sys.exception()` looks cleaner, but I don't think it's worth the extra noise for the next few years

vchudnov-g: Oh, that's a great idea. Let me try that.

vchudnov-g: OK, thanks. That worked. Merging.

PR #498, opened 2023-05-19
fix: add actionable errors for GCE long running operations

parthea:
Fixes b/280938007

Googlers see go/surface-actionable-errors-lro-gce

When an error occurs during an extended operation when using the python compute client, currently users only receive the error code and the error message, which is not helpful.
```
google.api_core.exceptions.BadRequest: 400 BAD REQUEST
```

The proposed solution is to provide users with actionable error information, surfacing the errors field which contains actionable errors. 
```
google.api_core.exceptions.BadRequest: 400 BAD REQUEST
INCOMPATIBLE_BACKEND_SERVICES: Validation failed for instance group 'projects/<REDACTED>/zones/us-central1-a/instanceGroups/my-instance-group-1': backend services 'projects/<REDACTED>/global/backendServices/my-backend-service-1' and 'projects/<REDACTED>global/backendServices/my-backend-service-2' point to the same instance group but the backends have incompatible balancing_mode. Values should be the same.
```


parthea (commented):
I'll add tests if go/surface-actionable-errors-lro-gce is approved.

vchudnov-g (approved):

PR #526, opened 2023-09-06
feat: add caching to routing header calculation

daniel-sanche:
_to_grpc_metadata is used to convert a tuple of string metadata into the proper  GCP routing format. It is part of the hot-path of rpcs with routing info, and it adds a noticable cost to each rpc. But typically, the same values will be used for each rpc

This PR adds a functools.lru_cache wrapper, to memoize the most recent paramaters used

```
import timeit
import functools
from google.api_core.gapic_v1.routing_header import to_grpc_metadata

metadata = (("table_name", "projects/p/tables/my-table"),)
result = timeit.timeit(lambda: to_grpc_metadata(metadata))
print(f"orig time: {result:0.2f}")

cached_to_metadata = functools.lru_cache(to_grpc_metadata)
result = timeit.timeit(lambda: cached_to_metadata(metadata))
print(f"cached time: {result:0.2f}")
```

```
orig time: 1.19
cached time: 0.07
```


parthea (approved):
Added minor observations, otherwise LGTM

PR #529, opened 2023-09-08
chore: avoid checking instance on each stream call

daniel-sanche:
_wrap_stream_errors currently tests the type of the call on each call. This PR moves the logic to happen once, at method wrap time.


parthea:
Converting to draft as presubmits are failing

daniel-sanche:
@parthea I fixed the tests. 

I'm prioritizing some of these gapic performance improvements, so I'm fine if you'd rather close this for now and revisit it later. But this one should be fairly straight-forward

parthea (approved):
LGTM, but please wait for @vchudnov-g to review

parthea:
@vchudnov-g Please could you review?

google/api_core/grpc_helpers_async.py
```diff
@@ -207,10 +196,17 @@ def wrap_errors(callable_):
...
         return _wrap_unary_errors(callable_)
+    elif isinstance(callable_, aio.UnaryStreamMultiCallable):
+        return _wrap_stream_errors(callable_, _WrappedUnaryStreamCall)
+    elif isinstance(callable_, aio.StreamUnaryMultiCallable):
+        return _wrap_stream_errors(callable_, _WrappedStreamUnaryCall)
+    elif isinstance(callable_, aio.StreamStreamMultiCallable):
+        return _wrap_stream_errors(callable_, _WrappedStreamStreamCall)
     else:
-        return _wrap_stream_errors(callable_)
+        raise TypeError("Unexpected type of callable: {}".format(type(callable_)))
```
vchudnov-g, line 209: nit (not a blocker, consider for the future): consider in cases like this raising a custom subclass of `TypeError`, so that we can check the exception type in tests

vchudnov-g (approved):
Thanks for doing this!

PR #541, opened 2023-10-21
fix: drop usage of distutils

### LINKED ISSUE #507: Remove dependency on deprecated distutils
The distutils module has been deprecated and will be removed in Python3.12: https://docs.python.org/3.11/library/distutils.html

setuptools also discourages using its vendored version: https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html#prefer-setuptools

The explicit `distutils` dependency / import in `abstract_operations_client.py` is only needed for one call to `distutils.util.strtobool()`: https://github.com/googleapis/python-api-core/blob/7dfc3a7a439243f05238a11b68a31720fde1769e/google/api_core/operations_v1/abstract_operations_client.py#L297

We encountered this under Python3.10 due to the way we build / run our code under Docker. This can be easily circumvented in our case by installing the `python3-distutils` Debian package, but the dependency should probably be removed from this package.

#### Environment details

  - OS type and version: Linux 218300304P 6.1.0-9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.27-1 (2023-05-08) x86_64 GNU/Linux
  - Python version: Python 3.12.0b2
  - pip version: pip 23.1.2 from /tmp/venv/lib/python3.12/site-packages/pip (python 3.12)
  - `google-api-core` version: ```Name: google-api-core
Version: 2.11.0
Summary: Google API client core library
Home-page: https://github.com/googleapis/python-api-core
Author: Google LLC
Author-email: googleapis-packages@google.com
License: Apache 2.0
Location: /tmp/venv/lib/python3.12/site-packages
Requires: google-auth, googleapis-common-protos, protobuf, requests
Required-by:```

#### Steps to reproduce

  1. `python3.12 -m venv /tmp/venv`
  2. `/tmp/venv/bin/pip install google-api-core`

#### Code example

```python
from google.api_core.operations_v1.abstract_operations_client import AbstractOperationsClient
```

#### Stack trace
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/tmp/venv/lib/python3.12/site-packages/google/api_core/operations_v1/__init__.py", line 17, in <module>
    from google.api_core.operations_v1.abstract_operations_client import AbstractOperationsClient
  File "/tmp/venv/lib/python3.12/site-packages/google/api_core/operations_v1/abstract_operations_client.py", line 17, in <module>
    from distutils import util
ModuleNotFoundError: No module named 'distutils'
```

parthea:
Fixes https://github.com/googleapis/python-api-core/issues/507 🦕

- Raises `ValueError` if the format of the `GOOGLE_API_USE_CLIENT_CERTIFICATE` is not as defined below
https://github.com/googleapis/python-api-core/blob/405272c05f8c6d20e242c6172b01f78f0fd3bf32/google/api_core/operations_v1/abstract_operations_client.py#L275-L280

This is the same behaviour that exists in `gapic-generator-python` here https://github.com/googleapis/gapic-generator-python/blob/45f18a69454bcff68c0f98535c68fbdb4e09db1d/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/client.py.j2#L253-L256

### DIFF
--- a/google/api_core/operations_v1/abstract_operations_client.py
+++ b/google/api_core/operations_v1/abstract_operations_client.py
@@ -293,14 +293,14 @@ class AbstractOperationsClient(metaclass=AbstractOperationsClientMeta):
             client_options = client_options_lib.ClientOptions()
 
         # Create SSL credentials for mutual TLS if needed.
-        use_client_cert = os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE", "false")
-        if use_client_cert.lower() not in ("true", "false"):
+        use_client_cert = os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE", "false").lower()
+        if use_client_cert not in ("true", "false"):
             raise ValueError(
                 "Environment variable `GOOGLE_API_USE_CLIENT_CERTIFICATE` must be either `true` or `false`"
             )
         client_cert_source_func = None
         is_mtls = False
-        if use_client_cert.lower() == "true":
+        if use_client_cert == "true":
             if client_options.client_cert_source:
                 is_mtls = True
                 client_cert_source_func = client_options.client_cert_source
--- a/google/api_core/operations_v1/abstract_operations_client.py
+++ b/google/api_core/operations_v1/abstract_operations_client.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 #
 from collections import OrderedDict
-from distutils import util
 import os
 import re
 from typing import Dict, Optional, Sequence, Tuple, Type, Union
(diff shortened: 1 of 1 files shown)


bcoe (approved):

google/api_core/operations_v1/abstract_operations_client.py
```diff
@@ -294,13 +293,14 @@ def __init__(
             client_options = client_options_lib.ClientOptions()
 
         # Create SSL credentials for mutual TLS if needed.
-        use_client_cert = bool(
-            util.strtobool(os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE", "false"))
-        )
-
+        use_client_cert = os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE", "false")
```
vchudnov-g, line 296: Might as well us `lower()` in this assignment so you don't have to repeat it below

(It would be great to cast this to a bool like the previous code was doing, but it seems like too much overhead for the very limited use in a very compact LOC range. The string comparison is OK.)

vchudnov-g (approved):
Only a minor, non-blocking comment.

google/api_core/operations_v1/abstract_operations_client.py
parthea: Fixed in 614150d26ee14e697f0afab655da794b745bc99e

PR #543, opened 2023-10-26
fix: add warning to retry target to avoid incorrect usage

### LINKED ISSUE #542: Raise a warning in google.api_core.retry.retry_target if the return value is awaitable
A warning can be raised within `google.api_core.retry.retry_target` depending on whether the return value of `target()` is awaitable in the following way:

```python3
if inspect.isawaitable(target()):
            raise ValueError("Incorrect use of Retry in async code. Use AsyncRetry instead")
```

This will ensure that `google.api_core.retry.Retry` is not incorrectly used in async clients/code and will suggest users to use `google.api_core.retry_async.AsyncRetry` instead

related issue: [fix retries in async client](https://github.com/googleapis/google-cloud-python/issues/11405)

ohmayr:
Fixes #542 🦕

### DIFF
--- a/google/api_core/retry.py
+++ b/google/api_core/retry.py
@@ -62,6 +62,8 @@ import logging
 import random
 import sys
 import time
+import inspect
+import warnings
 from typing import Any, Callable, TypeVar, TYPE_CHECKING
 
 import requests.exceptions
@@ -84,6 +86,7 @@ _DEFAULT_INITIAL_DELAY = 1.0  # seconds
 _DEFAULT_MAXIMUM_DELAY = 60.0  # seconds
 _DEFAULT_DELAY_MULTIPLIER = 2.0
 _DEFAULT_DEADLINE = 60.0 * 2.0  # seconds
+_ASYNC_RETRY_WARNING = "Using the synchronous google.api_core.retry.Retry with asynchronous calls may lead to unexpected results. Please use google.api_core.retry_async.AsyncRetry instead."
 
 
 def if_exception_type(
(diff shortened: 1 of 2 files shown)


google/api_core/retry.py
```diff
@@ -201,7 +202,12 @@ def retry_target(
 
     for sleep in sleep_generator:
         try:
-            return target()
+            result = target()
+            if inspect.isawaitable(result):
+                raise exceptions.GoogleAPICallError(
+                    "Warning: Use google.api_core.retry_async.AsyncRetry for async calls!"
```
vchudnov-g, line 208: Consider changing the wording to something like
"Warning: Using the synchronous google.api_core.retry.Retry with asynchronmous calls may lead to unexpected results. Please use google.api_core.retry_async.AsyncRetry instead."

Motivation: we summarize the problem for the user, giving context for our warning/recommendation.

vchudnov-g (changes requested):
It's looking good, but just a few questions/comments.

tests/unit/test_retry.py
```diff
@@ -129,6 +129,25 @@ def test_retry_target_non_retryable_error(utcnow, sleep):
...
+    return_value=datetime.datetime.min,
+    autospec=True,
+)
+@pytest.mark.asyncio
+async def test_retry_target_warning_for_retry(utcnow, sleep):
+    predicate = retry.if_exception_type(ValueError)
+    target = mock.AsyncMock(spec=["__call__"])
+
+    with pytest.warns(Warning) as exc_info:
+        retry.retry_target(target, predicate, range(10), None)
```
vchudnov-g, line 144: IIUC, `predicate` is not used, right? So I would suggest one of (a) adding a comment in line 140 saying that this is just a filler that shouldn't affect the test, or (b) setting  it to `None`, or (c) doing each of (a) and (b) in separate test cases, to show `predicate` doesn't affect things.

I think just (a) would be fine, though.

vchudnov-g (approved):

PR #546, opened 2023-11-06
feat: add universe_domain argument to ClientOptions

### LINKED ISSUE #547: Feat: Add support for universe_domain as a client option for TPC
`universe_domain` is required as an optional field in `google.api_core.ClientOptions` to enable users to explicitly configure universe domain as a client option. 

This will enable API endpoint to be resolved using a templated value if a universe domain is specified or otherwise default to GDU.

```python3

# user configures the universe domain:
client = Client(google.api_core.ClientOptions{universe_domain="foo.com"})

# api_endpoint points to the TPC universe:
api_endpoint = "service.foo.com"

```

```python3

# user does not configure the universe domain:
client = Client(google.api_core.ClientOptions{})

# api_endpoint defaults to GDU:
api_endpoint = "service.googleapis.com"

```

ohmayr:
Fixes #547 🦕

BEGIN_COMMIT_OVERRIDE
feat: add universe_domain argument to ClientOptions 
END_COMMIT_OVERRIDE


google/api_core/client_options.py
```diff
@@ -91,6 +91,7 @@ def __init__(
         scopes=None,
         api_key=None,
         api_audience=None,
+        universe_domain=None,
```
parthea, line 94: Please update the docstring as well which starts at line 53

google/api_core/client_options.py
```diff
@@ -106,6 +112,7 @@ def __init__(
         self.scopes = scopes
         self.api_key = api_key
         self.api_audience = api_audience
+        self.universe_domain = universe_domain
```
arithmetic1728, line 115: should the logic here be:

self.universe_domain = universe_domain if universe_domain else googleapis.com

ohmayr: I'd like to disagree because:
1 - If the universe domain in the client options is empty, then we want to check if it's provided via an Env variable.
2 - The default value is handled in the client code itself by introducing `GOOGLE_DEFAULT_UNIVERSE = "googleapis.com"`

arithmetic1728: I see. Sounds good.

google/api_core/client_options.py
```diff
@@ -75,6 +75,11 @@ class ClientOptions(object):
             authentication flows. Audience is typically a resource identifier.
             If not set, the service endpoint value will be used as a default.
             An example of a valid ``api_audience`` is: "https://language.googleapis.com".
+        universe_domain (Optional[str]): The desired universe domain, must match the one in
+            credentials. If not set, the default universe domain is "googleapis.com".
+            If both api_endpoint and universe_domain are set, then api_endpoint is used
+            as the service endpoint.
+            format: {service}.{universe_domain}.
```
parthea, line 82:
```suggestion
        universe_domain (Optional[str]): The desired universe domain. This must match the one in
            credentials. If not set, the default universe domain is `googleapis.com`.
            If both `api_endpoint` and `universe_domain` are set, then `api_endpoint` is used
            as the service endpoint. If `api_endpoint` is not specified, the format will be
            `{service}.{universe_domain}`.
```

parthea (approved):

PR #550, opened 2023-11-09
feat: support with_call for wrapped rpcs

daniel-sanche:
Most grpc calls return a `grpc.Call` object, that contains trailing metadata and other useful information. You can then `await` (for async unary rpcs) or iterate over (for streaming rpcs) the Call object to get the proto results of the rpc.

This is not the case for unary synchronous calls, because there is no intermediate object. Calling the rpc results in the value directly. The grpc library provides [`with_call` methods](https://grpc.github.io/grpc/python/grpc.html#grpc.UnaryUnaryMultiCallable.with_call) for this reason, which return a tuple of the rpc result and the Call object.

Currently, the gapic wrapper does not provide a way to access this data

This PR solves the issue by adding `with_call` to the `wrap_method` function, which will is passed down to the grpc callable

### DIFF
--- a/google/api_core/gapic_v1/method.py
+++ b/google/api_core/gapic_v1/method.py
@@ -137,6 +137,8 @@ def wrap_method(
     default_timeout=None,
     default_compression=None,
     client_info=client_info.DEFAULT_CLIENT_INFO,
+    *,
+    with_call=False,
 ):
     """Wrap an RPC method with common behavior.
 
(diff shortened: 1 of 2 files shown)


vchudnov-g (approved):

vchudnov-g:
@daniel-sanche Could you fix the lint error?

daniel-sanche:
> @daniel-sanche Could you fix the lint error?

fixed, thanks

PR #554, opened 2023-11-17
feat: add type annotations to wrapped grpc calls

daniel-sanche:
Our gapic libraries declare that they return `Iterable[SomeProto]` or `Awaitable[SomeProto]` or `Awaitable[AsyncIterable[SomeProto]`, but that doesn't tell the whole story. The objects returned are actually[ `grpc.Call` subclasses](https://github.com/grpc/grpc/blob/0a71e076f8185347be3a27fe8a64cd6c8ecc0d09/src/python/grpcio/grpc/aio/_call.py#L181), which expose the ability to retrieve metadata from the rpc, or call `cancel`, or other trigger useful functionality. 

Unfortunately, there is currently no way to use these `grpc.Call` methods without mypy errors, because of the restrictive return type annotation used

This PR is the first step in addressing this, by giving us a more powerful return type in `api-core`.

On the sync side, I made _StreamingResponseIterator into a Generic container, and gave it a better public facing name. This way, we we can return `GrpcStream[SomeProto]` instead of `Iterable[SomeProto]`, with all grpc.Call methods accessible

On the Async side, I made every `_WrappedXYResponse` class into a Generic container, and declared new `GrpcAsyncStream[SomeProto]` and `AwaitableGrpcCall[SomeProto]` types that can be used in place of `AsyncIterable[SomeProto]` and `Awaitable[SomeProto]` respectively.

For more context on the motivating problem, see https://github.com/googleapis/gapic-generator-python/issues/1856. This PR lays some of the groundwork for resolving that issue in the future. But the type improvements here should also stand alone

### DIFF
--- a/google/api_core/grpc_helpers.py
+++ b/google/api_core/grpc_helpers.py
@@ -58,6 +58,7 @@ _LOGGER = logging.getLogger(__name__)
 # denotes the type yielded from streaming calls
 S = TypeVar("S")
 
+
 def _patch_callable_name(callable_):
     """Fix-up gRPC callable attributes.
 
--- a/google/api_core/grpc_helpers.py
+++ b/google/api_core/grpc_helpers.py
@@ -148,6 +148,7 @@ class _StreamingResponseIterator(grpc.Call, Generic[S]):
         return self._wrapped.trailing_metadata()
 
 
+# public type alias denoting the return type of streaming gapic calls
 GrpcStream = _StreamingResponseIterator[S]
 
 
--- a/google/api_core/grpc_helpers.py
+++ b/google/api_core/grpc_helpers.py
@@ -58,6 +58,7 @@ _LOGGER = logging.getLogger(__name__)
 # denotes the type yielded from streaming calls
 S = TypeVar("S")
 
+
 def _patch_callable_name(callable_):
     """Fix-up gRPC callable attributes.
 
@@ -165,9 +166,7 @@ def _wrap_stream_errors(callable_):
             # hidden flag to see if pre-fetching is disabled.
             # https://github.com/googleapis/python-pubsub/issues/93#issuecomment-630762257
             prefetch_first = getattr(callable_, "_prefetch_first_result_", True)
-            return GrpcStream(
-                result, prefetch_first_result=prefetch_first
-            )
+            return GrpcStream(result, prefetch_first_result=prefetch_first)
         except grpc.RpcError as exc:
             raise exceptions.from_grpc_error(exc) from exc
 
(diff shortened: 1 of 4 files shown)


parthea:
@daniel-sanche Please could you address the lint failure?

daniel-sanche:
@parthea done

google/api_core/grpc_helpers_async.py
```diff
@@ -127,26 +134,32 @@ async def done_writing(self):
...
-    _WrappedStreamRequestMixin, _WrappedStreamResponseMixin, aio.StreamStreamCall
+    _WrappedStreamRequestMixin, _WrappedStreamResponseMixin[S], aio.StreamStreamCall
 ):
     """Wrapped StreamStreamCall to map exceptions."""
 
 
+# public type alias denoting the return type of streaming gapic calls
+GrpcAsyncStream = _WrappedStreamResponseMixin[S]
+# public type alias denoting the return type of unary gapic calls
+AwaitableGrpcCall = _WrappedUnaryResponseMixin[U]
```
daniel-sanche, line 160: Let me know if you have other naming suggestions for these (`AsyncGrpcCall`? `GrpcAsyncIterable`?)

tests/asyncio/test_grpc_helpers_async.py
```diff
@@ -266,6 +266,28 @@ def test_wrap_errors_non_streaming(wrap_unary_errors):
     wrap_unary_errors.assert_called_once_with(callable_)
 
 
+def test_grpc_async_stream():
+    """
+    GrpcAsyncStream type be both an AsyncIterator and a grpc.aio.Call.
```
parthea, line 271:
```suggestion
    GrpcAsyncStream type should be both an AsyncIterator and a grpc.aio.Call.
```

google/api_core/grpc_helpers.py
```diff
@@ -79,7 +83,7 @@ def error_remapped_callable(*args, **kwargs):
     return error_remapped_callable
 
 
-class _StreamingResponseIterator(grpc.Call):
+class GrpcStream(grpc.Call, Generic[S]):
```
parthea, line 86: Do you think this could be a breaking change? Could we keep the name as `_StreamingResponseIterator` ? There are many hits for `_StreamingResponseIterator` in Google search. I'm worried that changing `_StreamingResponseIterator` could cause issues downstream in user code as it is a response we provide.

daniel-sanche: Makes sense, I'll create a type alias instead, like I did on the async side

parthea (approved):

parthea (approved):

PR #592, opened 2024-01-31
fix: Retry constructors methods support None

### LINKED ISSUE #591: API ignores user defined timeout and applies default timeout
When we updated to `dbt` version 1.7.5 and `dbt-bigquery` version 1.7.3, we noticed that our BigQuery queries were always timing out and cancelling after 900 seconds with the below error message, regardless of our own timeout settings:
```
Query exceeded configured timeout of 5400s
```
Those two `dbt` libs depend on the `google-api-core` version 2.16.0, which seems to always use the default timeout of 900 seconds instead of honouring the user defined timeout setting. This appears to be happening somewhere here: https://github.com/googleapis/python-api-core/blob/v2.16.0/google/api_core/future/polling.py#L126

Once we reverted to using `google-api-core` version 2.15.0, this problem disappeared and dbt went back to honouring our custom defined timeout settings.

daniel-sanche:
Fixes https://github.com/googleapis/python-api-core/issues/591

v2.16.0 did some refactoring of the Retry class, including adding a `_replace` helper method to simplify the constructors `with_deadline`, `with_timeout`, `with_predicate`, and `with_delay`. But it was noticed there is a bug in the implementation, where falsy values are ignored, with the existing value silently used instead.

Note that `None` is technically not supported by the type annotations, and 0 is typically not a valid value for these fields, which is likely why there was a gap in the tests around this. But the annotations were added recently, and existing code use pass None in for a timeouts, and None is supported by the internal retry code down the stack. For this reason, I changed the type annotation for the timeout field to support None as part of this fix, and added tests to ensure that it will be supported in the future


google/api_core/retry/retry_base.py
```diff
@@ -318,26 +309,42 @@ def with_predicate(self, predicate: Callable[[Exception], bool]) -> Self:
         Returns:
             Retry: A new retry instance with the given predicate.
         """
-        return self._replace(predicate=predicate)
+        return type(self)(
+            predicate=predicate,
```
vchudnov-g, line 313: Are we allowing `predicate` to be `None` here? If so, we would need to tweak the type annotation, and to add a test for this edge case? (The test might be starting with something with a non-trivial `predicate`, and then verifying that calling `.with_predicate(None)` sets it to `None`?

google/api_core/retry/retry_base.py
```diff
@@ -318,26 +309,42 @@ def with_predicate(self, predicate: Callable[[Exception], bool]) -> Self:
...
+                value is used.
 
         Returns:
-            Retry: A new retry instance with the given predicate.
+            Retry: A new retry instance with the given delay options.
         """
-        return self._replace(initial=initial, maximum=maximum, multiplier=multiplier)
+        return type(self)(
+            predicate=self._predicate,
+            initial=initial if initial is not None else self._initial,
```
vchudnov-g, line 342: If these values are `not None` but still falsy, we still want to use them? If so, we should capture this in a test since this differs from the previous behavior.

vchudnov-g (approved):
Some questions and test suggestions, but the core functionality LGTM.

google/api_core/retry/retry_base.py
daniel-sanche: It seems like predicate has always been a required variable

google/api_core/retry/retry_base.py
daniel-sanche: Yeah it differs from the previous behaviour, but it is in line with the [pre-2.16.0 behaviour](https://github.com/googleapis/python-api-core/blob/9866183b42f632dddade9ba6c69f9b3015118dd4/google/api_core/retry.py#L457)

Good point on the test. I had added one for timeout, but not for this one. Fixed

parthea: @daniel-sanche  Please could you link the specific commit for the fix in https://github.com/googleapis/python-api-core/pull/592#discussion_r1476827410?

daniel-sanche: sorry, forgot to push. The new test should be there now

google/api_core/retry/retry_base.py
vchudnov-g: Well, in `replace`, we had `predicate: Callable[[Exception], bool] | None = None,`, right?

daniel-sanche: Sorry, I missed this comment

In the _replace helper, `None` represented a variable that isn't replaced (that is, it used the same value as the original Retry object). 

But this  doesn't work if users are passing in `None` for a `with_timeout` and expecting it to disable timeouts, which wasn't supported by the type annotations, but was used in practice. Which was the motivation for this change

PR #621, opened 2024-02-27
feat: add common logic for supporting universe domain

### LINKED ISSUE #624: Add static code for supporting universe domain (to be used in both GAPICs and Apiary)
Universes other than the default `"googleapis.com"` need to be supported in both GAPICs and Apiary. The following code was added to GAPICs but can be moved to `python-api-core`:

```python3
def _get_universe_domain(
    client_universe_domain: Optional[str], universe_domain_env: Optional[str]
) -> str:  # temp disable Optional
    """Return the universe domain used by the client.

    Args:
        client_universe_domain (Optional[str]): The universe domain configured via the client options.
        universe_domain_env (Optional[str]): The universe domain configured via the "GOOGLE_CLOUD_UNIVERSE_DOMAIN" environment variable.

    Returns:
        str: The universe domain to be used by the client.

    Raises:
        ValueError: If the universe domain is an empty string.
    """
    universe_domain = _DEFAULT_UNIVERSE
    if client_universe_domain is not None:
        universe_domain = client_universe_domain
    elif universe_domain_env is not None:
        universe_domain = universe_domain_env
    if len(universe_domain.strip()) == 0:
        raise _Empty_Universe_Error
    return universe_domain


def _compare_universes(
    client_universe: str, credentials: Union[ga_credentials, oauth2_credentials]
) -> bool:
    """Returns True iff the universe domains used by the client and credentials match.

    Args:
        client_universe (str): The universe domain configured via the client options.
        credentials Union[google.auth.credentials.Credentials, oauth2client.client.Credentials]: The credentials being used in the client.

    Returns:
        bool: True iff client_universe matches the universe in credentials.

    Raises:
        ValueError: when client_universe does not match the universe in credentials.
    """
    credentials_universe = getattr(credentials, "universe_domain", _DEFAULT_UNIVERSE)

    if client_universe != credentials_universe:
        raise _UniverseMismatchError(client_universe, credentials_universe)
    return True

```

ohmayr:
Fixes #624 🦕


google/api_core/universe_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def _get_universe_domain(
+    client_universe_domain: Optional[str], universe_domain_env: Optional[str]
+) -> str:
+    """Return the universe domain used by the client.
+
+    Args:
+        client_universe_domain (Optional[str]): The universe domain configured via the client options.
+        universe_domain_env (Optional[str]): The universe domain configured via the "GOOGLE_CLOUD_UNIVERSE_DOMAIN" environment variable.
```
vchudnov-g, line 45: nit: line length

vchudnov-g, line 45: Where do we read the environment variable? Ideally, that should probably be one of the helpers too. WDYT?

(And there, potentially, we could put the name of this environment variable into a constant and possibly refer to that constant here. I like the generality of that, but since  that would be another layer of indirection for developers reading the code, maybe not.)

google/api_core/universe_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+        message = (
+            f"The configured universe domain ({client_universe}) does not match the universe domain "
+            f"found in the credentials ({credentials_universe}). "
+            "If you haven't configured the universe domain explicitly, "
+            f"`{_DEFAULT_UNIVERSE}` is the default."
+        )
+        super().__init__(message)
+
+
+def _get_universe_domain(
```
vchudnov-g, line 38: These two methods are underscore private, but we want to make them public here, right?

google/api_core/universe_helpers.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2023 Google LLC
```
vchudnov-g, line 1: Date

tests/unit/test_universe_helpers.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2023 Google LLC
```
vchudnov-g, line 1: Date

vchudnov-g (commented):
Looking good, but I left some comments.

One other comment: do you want to call this module `universe_helpers`?

```
from google.api_core import  universe_helpers

universe_helpers.get_universe_domain()
universe_helpers.compare_universes()
```

I wonder whether there are better names for this module and methods.

Maybe `universe_domain.determine()` and `universe_domain.compare()`? Or maybe we need slightly more explicit names than that (`universe.determine_domain()`, `universe.compare_domains()`?)

(Feel free to push back)

tests/unit/test_universe.py
```diff
@@ -0,0 +1,49 @@
...
+def test_determine_domain():
+    client_universe_domain = "foo.com"
+    universe_domain_env = "bar.com"
+
+    assert (
+        universe.determine_domain(client_universe_domain, universe_domain_env)
+        == client_universe_domain
+    )
+    assert universe.determine_domain(None, universe_domain_env) == universe_domain_env
+    assert universe.determine_domain(None, None) == universe.DEFAULT_UNIVERSE
```
vchudnov-g, line 29: For completeness, maybe we should also test the cases `(client_universe_domain, None)` and `(None, "")`

tests/unit/test_universe.py
```diff
@@ -0,0 +1,49 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from google.api_core import universe
+from google.auth import credentials
+
+
+def test_determine_domain():
+    client_universe_domain = "foo.com"
```
vchudnov-g, line 21: nit: the naming of these two variables is inconsistent. How about `domain_client`, `domain_env`?

tests/unit/test_universe.py
```diff
@@ -0,0 +1,49 @@
...
+        "found in the credentials (googleapis.com). "
+        "If you haven't configured the universe domain explicitly, "
+        "`googleapis.com` is the default."
+    )
+
+    assert universe.compare_domains(universe.DEFAULT_UNIVERSE, ga_credentials) is True
+
+    with pytest.raises(ValueError) as excinfo:
+        universe.compare_domains("foo.com", ga_credentials)
+    assert str(excinfo.value) == mismatch_err_msg
```
vchudnov-g, line 49: Let's not compare the full text of the exception here; this is a brittle pattern, since we may choose to fine tune the text in the future.

Let's just change the code above to `pytest.raises(_UniverseMismatchError)`. If you want, you could check that the text _contains_ the values of `client_universe` and `credentials_universe`

tests/unit/test_universe.py
```diff
@@ -0,0 +1,49 @@
...
+    assert universe.determine_domain(None, universe_domain_env) == universe_domain_env
+    assert universe.determine_domain(None, None) == universe.DEFAULT_UNIVERSE
+
+    with pytest.raises(ValueError) as excinfo:
+        universe.determine_domain("", None)
+    assert str(excinfo.value) == "Universe Domain cannot be an empty string."
+
+
+def test_compare_domains():
+    ga_credentials = credentials.AnonymousCredentials()
```
vchudnov-g, line 37: You could generalize the test, and make it not depend on auth, by creating an object with a `universe_domain` attribute. This would also allow you to test cases where that value is different from `universe.DEFAULT_UNIVERSE`).

And you could also use an object without a `universe_domain` attr, to ensure the default gets set corectly.

google/api_core/universe.py
```diff
@@ -0,0 +1,81 @@
...
+from typing import Any, Optional
+
+DEFAULT_UNIVERSE = "googleapis.com"
+
+mTLS_Universe_Error = MutualTLSChannelError(
+    f"mTLS is not supported in any universe other than {DEFAULT_UNIVERSE}."
+)
+
+
+class _UniverseMismatchError(ValueError):
```
vchudnov-g, line 27: Why keep this private? If we make it public, the user can check whether exceptions they get are of this type.

tests/unit/test_universe.py
```diff
@@ -0,0 +1,54 @@
...
+    fake_domain = "foo.com"
+
+    assert (
+        universe.compare_domains(universe.DEFAULT_UNIVERSE, _Fake_Credentials()) is True
+    )
+    assert universe.compare_domains(fake_domain, _Fake_Credentials(fake_domain)) is True
+
+    with pytest.raises(universe.UniverseMismatchError) as excinfo:
+        universe.compare_domains(
+            universe.DEFAULT_UNIVERSE, _Fake_Credentials(fake_domain)
```
vchudnov-g, line 51: And how about two more cases with `(fake_domain, _Fake_Credentials())` and `(fake_domain, _Fake_Credentials(another_fake_domain))`?

google/api_core/universe.py
```diff
@@ -0,0 +1,87 @@
...
+# limitations under the License.
+
+"""Helpers for universe domain."""
+
+from google.auth.exceptions import MutualTLSChannelError
+from typing import Any, Optional
+
+DEFAULT_UNIVERSE = "googleapis.com"
+
+mTLS_Universe_Error = MutualTLSChannelError(
```
vchudnov-g, line 22: Do we test for this anywhere? We should. Possibly in a separate PR if it's not relevant here....but then maybe we should omit this declaration, since it's not being used anywhere in this PR.

vchudnov-g (commented):
Could you either use or remove the `mTLS_Universe_Error` declaration from this PR?. Right now it's not being used anywhere.

Aside from that, just a minor (non-blocking) testing suggestion.

google/api_core/universe.py
ohmayr: addressed in f5bff28d6d2deda1a7316a065e3752cd0c5bed73.

tests/unit/test_universe.py
ohmayr: addressed in f5bff28d6d2deda1a7316a065e3752cd0c5bed73.

tests/unit/test_universe.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+    assert str(excinfo.value).find(universe.DEFAULT_UNIVERSE) >= 0
+    assert str(excinfo.value).find(fake_domain) >= 0
+
+    with pytest.raises(universe.UniverseMismatchError) as excinfo:
+        universe.compare_domains(fake_domain, _Fake_Credentials())
```
vchudnov-g, line 58: These won't fit in one line? Darn.

tests/unit/test_universe.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_compare_domains():
+    fake_domain = "foo.com"
+    another_fake_domain = "bar.com"
+
+    assert (
+        universe.compare_domains(universe.DEFAULT_UNIVERSE, _Fake_Credentials()) is True
+    )
+    assert universe.compare_domains(fake_domain, _Fake_Credentials(fake_domain)) is True
```
vchudnov-g, line 48: Here and above, you shouldn't need `is True`, right? You should just be able to `assert function_returning_boolean(...)`

vchudnov-g (approved):

PR #626, opened 2024-02-29
fix(deps): Require proto-plus >= 1.22.3

### LINKED ISSUE #625: Undeclared dependency on proto-plus
In https://github.com/googleapis/python-api-core/pull/604, `import proto` was added to `google/api_core/rest_streaming.py` however the dependency was not added to `setup.py`. See https://github.com/googleapis/python-api-core/commit/bcebc92eca69dae81c5e546d526c92b164a6b3b4#r139197223 where the issue was reported.

https://github.com/googleapis/python-api-core/blob/475ff2efa26f1b6249f07d1781c1324685fe23dd/google/api_core/rest_streaming.py#L21

parthea:
Fixes #625 🦕

Thanks @shadchin for reporting the issue!

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -44,16 +44,6 @@ nox.options.sessions = [
 ]
 
 
-def _greater_or_equal_than_37(version_string):
-    tokens = version_string.split(".")
-    for i, token in enumerate(tokens):
-        try:
-            tokens[i] = int(token)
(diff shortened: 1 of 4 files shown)


vchudnov-g (approved):
Thanks for fixing this and for removing the stale checks for unsupported language versions!

PR #638, opened 2024-04-08
feat: add google.api_core.version_header

parthea:
Towards b/330968465

### DIFF
--- a/google/api_core/version_header.py
+++ b/google/api_core/version_header.py
@@ -1,0 +1,29 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


ohmayr (approved):

google/api_core/version_header.py
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+API_VERSION_METADATA_KEY = "x-goog-api-version"
+
+
+def to_api_version_header(version_identifier):
+    """Returns the API Version header for the given `version_identifier`.
```
vchudnov-g, line 19: nit: it's not the header itself, but the data for the header....

vchudnov-g (approved):
Thanks for doing this!

google/api_core/version_header.py
parthea: Fixed in 60b33e60b795715cca9df43edd2d8ed28cd5353c

PR #680, opened 2024-07-29
fix: Fail gracefully if could not import `rpc_status` module

### LINKED ISSUE #659: AttributeError: 'NoneType' object has no attribute 'from_call'
In `exceptions.py`, if `grpcio-status` is not installed, `rpc_status` is set to `None`

https://github.com/googleapis/python-api-core/blob/d96eb5cdd8120bfec97d62b09512c6fecc325be8/google/api_core/exceptions.py#L31-L43

In `_parse_grpc_error_details`, when `rpc_status` is `None`, this leads to `AttributeError: 'NoneType' object has no attribute 'from_call'`
https://github.com/googleapis/python-api-core/blob/d96eb5cdd8120bfec97d62b09512c6fecc325be8/google/api_core/exceptions.py#L562-L566

We should add a check to make sure `rpc_status` is not `None` before calling `rpc_status.from_call(rpc_exc)`

NickGoog:
Fail gracefully if could not import `rpc_status` module.

Fixes #659 🦕


parthea:
I'm looking into the failing coverage check
https://github.com/googleapis/python-api-core/actions/runs/10200237862/job/28219271887

parthea (approved):

PR #688, opened 2024-08-22
feat: Add support for creating exceptions from an asynchronous response

ohmayr:
~~This PR implements a `_RestCall` class to map `GoogleAPICallError` to an asynchronous rest callable that raises `google.auth.exceptions.TransportError`.This wrapping is done via a helper method `rest_helpers_async.wrap_errors`.~~

A new parameter `kind` is introduced in `wrap_method` which can be one of `grpc_asyncio` (Default Value) or `rest_asyncio` to distinguish between the type of callables so that we can apply the appropriate wrapping logic.

The property kind is exposed at the transport level [here](https://github.com/googleapis/gapic-generator-python/blob/f773c8a8634c1c1205ffa324da8f7d56c5462e37/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/base.py.j2#L357) and will be configured when wrapping methods during the transport initialization [here](https://github.com/googleapis/gapic-generator-python/blob/f773c8a8634c1c1205ffa324da8f7d56c5462e37/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/base.py.j2#L150).

As a follow up, a similar logic must be implemented for synchronous rest callables to avoid gRPC specific stack trace within a rest call.

### RE: Updated Workflow:

I've removed the proposed `_RestCall` class after determining that it is not needed to wrap a rest callable and just adds unnecessary logic.

The new proposed workflow is as follows:
- This PR still introduces a new parameter `kind` in `wrap_method` to determine the type of a callable i.e. `grpc_asyncio` or `rest_asyncio`.
- If `kind=="rest_asyncio"`, we skip the call to `grpc_helpers_async.wrap_errors` which is gRPC specific.

Essentially, this does the trick if we follow a similar pattern for error handling to what we're doing in a sync rest call. 

We have been lazily creating and raising a `GoogleAPICallError` from a `requests.Response` within our GAPICs using the [from_http_response](https://github.com/googleapis/python-api-core/blob/b5ddd6236560f70405c0ea8614d67b27230341c5/google/api_core/exceptions.py#L479) method [here](https://github.com/googleapis/gapic-generator-python/blob/f773c8a8634c1c1205ffa324da8f7d56c5462e37/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/rest.py.j2#L416). Therefore, we don't need to add the error handling logic to the callable beforehand (we're currently doing both the things for a sync rest call and should skip wrapping the callable beforehand there too).

This PR only handles the error handling logic for an async REST call. It also introduces a new helper i.e. `format_http_response_error` to create and raise a `GoogleAPICallError` for a more generic response since the existing `from_http_response` is specific to handling a `requests` response.

#### Alternate approach for the introduced helper:

The proposed implementation of `format_http_response_error` takes in payload as an argument so that we don't need to await for it within the call and keep this function synchronous.

Alternatively, we can have an async version of `from_http_response` that can be used to create `GoogleAPICallError` for async rest calls.

#### Follow Up:
As a follow up, we should skip wrapping sync rest callable with `grpc_helpers.wrap_errors` (similar to what we're doing here) to avoid the unnecessary gRPC specific logic within a rest call stracktrace.


google/api_core/exceptions.py
```diff
@@ -476,22 +476,31 @@ def from_http_status(status_code, message, **kwargs):
...
+    method = method.upper()
+    message = "{method} {url}: {error}".format(
+        method=method,
+        url=url,
+        error=error,
+    )
+    return message
+
+
+def format_http_response_error(response, method, url, payload=None):
```
parthea, line 489: nit: add type hints 
nit: add code comments to clarify the reason that we're moving away from `from_http_response`

google/api_core/exceptions.py
```diff
@@ -476,22 +476,31 @@ def from_http_status(status_code, message, **kwargs):
     return error
 
 
-def from_http_response(response):
-    """Create a :class:`GoogleAPICallError` from a :class:`requests.Response`.
+def _format_error_message(error, method, url):
+    method = method.upper()
+    message = "{method} {url}: {error}".format(
+        method=method,
+        url=url,
```
vchudnov-g, line 483: These parameters are REST-specific, so maybe the name of this function should include `_rest_` to make that explicit.

google/api_core/exceptions.py
vchudnov-g, line 1: IIUC, the changes in this file are just refactoring?

google/api_core/gapic_v1/method_async.py
```diff
@@ -32,6 +32,7 @@ def wrap_method(
     default_timeout=None,
     default_compression=None,
     client_info=client_info.DEFAULT_CLIENT_INFO,
+    kind="grpc",
```
vchudnov-g, line 35: OK, and when we call `wrap_method` from the GAPIC, it will specify this parameter, right? So that will be in a follow-up PR in the generator repo?

vchudnov-g (commented):
Looks good. Just minor comments/questions. Please address @parthea's comments as well.

google/api_core/exceptions.py
ohmayr: Yes and `format_http_response_error` will be used to format an asynchronous response error and will aslso eventually replace `from_http_response` in a synchronous code path.

google/api_core/gapic_v1/method_async.py
ohmayr: That's correct! Note to myself to update the default value to `grpc_asyncio`.

ohmayr:
Adding a `do not merge` label until open comments are addressed and the default value is updated to `grpc_asyncio`.

google/api_core/exceptions.py
ohmayr: addressed.

google/api_core/gapic_v1/method_async.py
ohmayr: addressed.

google/api_core/exceptions.py
ohmayr: addressed with a comment.

daniel-sanche:
Would this fix the issue in https://github.com/googleapis/gapic-generator-python/pull/1764? Or is this something else?

google/api_core/gapic_v1/method_async.py
vchudnov-g: Update the PR description with the right default value.

vchudnov-g (approved):

ohmayr:
> googleapis/gapic-generator-python#1764

This PR removes grpc code path from the stack trace when an error is raised from an asynchronous rest call.

PR #694, opened 2024-09-10
feat: add async unsupported paramater exception

ohmayr:
This PR implements a new exception type i.e. `AsyncRestUnsupportedParameterError` which is raised when an unsupported parameter is configured against the async rest transport.

### DIFF
--- a/google/api_core/exceptions.py
+++ b/google/api_core/exceptions.py
@@ -444,7 +444,6 @@ class DeadlineExceeded(GatewayTimeout):
 
 class AsyncRestUnsupportedParameterError(NotImplementedError):
     """Raised when an unsupported parameter is configured against async rest transport."""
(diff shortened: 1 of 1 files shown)


google/api_core/exceptions.py
```diff
@@ -442,6 +442,12 @@ class DeadlineExceeded(GatewayTimeout):
     grpc_status_code = grpc.StatusCode.DEADLINE_EXCEEDED if grpc is not None else None
 
 
+class AsyncRestUnsupportedParameterError(NotImplementedError):
+    """Raised when an unsupported parameter is configured against async rest transport."""
+
```
vchudnov-g, line 447:
```suggestion
```

vchudnov-g (approved):

PR #701, opened 2024-09-20
chore: add `async_rest` extra for async rest dependencies

parthea:
This PR adds an extra for `async_rest` which captures the minimum version of `google-auth` which is required to use the `async_rest` feature.

### DIFF
--- a/google/api_core/rest_streaming_async.py
+++ b/google/api_core/rest_streaming_async.py
@@ -22,7 +22,9 @@ try:
     import google.auth.aio.transport
 except ImportError as e:  # pragma: NO COVER
     raise ImportError(
-        "google-api-core[async_rest] is required to use asynchronous rest streaming."
+        "`google-api-core[async_rest]` is required to use asynchronous rest streaming. "
+        "Install the `async_rest` extra of `google-api-core` using "
+        "`pip install google-api-core[async_rest]`."
     ) from e
 
 import google.protobuf.message
--- a/google/api_core/rest_streaming_async.py
+++ b/google/api_core/rest_streaming_async.py
@@ -22,7 +22,7 @@ try:
     import google.auth.aio.transport
 except ImportError as e:  # pragma: NO COVER
     raise ImportError(
-        "google-auth>=2.35.0 is required to use asynchronous rest streaming."
+        "google-api-core[async_rest] is required to use asynchronous rest streaming."
     ) from e
 
 import google.protobuf.message
--- a/noxfile.py
+++ b/noxfile.py
@@ -147,9 +147,7 @@ def default(session, install_grpc=True, prerelease=False, install_async_rest=Fal
             f"{constraints_dir}/constraints-{constraints_type}{PYTHON_VERSIONS[0]}.txt",
         )
         # This *must* be the last install command to get the package from source.
-        session.install(
-            "-e", f".[{','.join(install_extras)}]", "--no-deps"
-        )
+        session.install("-e", f".[{','.join(install_extras)}]", "--no-deps")
     else:
         constraints_file = (
             f"{constraints_dir}/constraints-{constraints_type}{session.python}.txt"
--- a/noxfile.py
+++ b/noxfile.py
@@ -38,7 +38,7 @@ nox.options.sessions = [
     "unit",
     "unit_grpc_gcp",
     "unit_wo_grpc",
-    "unit_with_auth_aio",
+    "unit_with_async_rest_extra",
     "cover",
     "pytype",
     "mypy",
@@ -110,7 +110,7 @@ def install_prerelease_dependencies(session, constraints_path):
         session.install(*other_deps)
 
 
-def default(session, install_grpc=True, prerelease=False, install_auth_aio=False):
+def default(session, install_grpc=True, prerelease=False, install_async_rest=False):
     """Default unit test session.
 
     This is intended to be run **without** an interpreter set, so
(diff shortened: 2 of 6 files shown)


google/api_core/rest_streaming_async.py
```diff
@@ -22,7 +22,7 @@
     import google.auth.aio.transport
 except ImportError as e:  # pragma: NO COVER
     raise ImportError(
-        "google-auth>=2.35.0 is required to use asynchronous rest streaming."
+        "google-api-core[async_rest] is required to use asynchronous rest streaming."
```
ohmayr, line 25: Shall we update this to:

```
"google-api-core[async_rest] is required to use asynchronous rest streaming." Install google-api-core with async_rest extra using `pip install google-api-core[async_rest]`."
```

noxfile.py
```diff
@@ -38,7 +38,7 @@
     "unit",
     "unit_grpc_gcp",
     "unit_wo_grpc",
-    "unit_with_auth_aio",
+    "unit_with_async_rest_extra",
```
ohmayr, line 41: `"unit_w_async_rest_extra"`

testing/constraints-3.7.txt
```diff
@@ -9,7 +9,6 @@ googleapis-common-protos==1.56.2
 protobuf==3.19.5
 google-auth==2.14.1
 requests==2.18.0
-packaging==14.3
```
ohmayr, line 12: why do we remove this?

testing/constraints-async-rest-3.7.txt
```diff
@@ -0,0 +1,17 @@
+# This constraints file is used to check that lower bounds
```
ohmayr, line 1: Why is a new file needed for async rest? Why not simply test against the existing constraints file? Or am I missing something here?

google/api_core/rest_streaming_async.py
parthea: Fixed in https://github.com/googleapis/python-api-core/pull/701/commits/a8af7912063eeaa09688051da2659a9badab9d08

noxfile.py
parthea: Fixed in https://github.com/googleapis/python-api-core/pull/701/commits/9f7d36c90ed2f4f287a9305e6b18431f04d616c6

testing/constraints-3.7.txt
parthea: `packaging` is not a direct dependency. It doesn't need to be in `testing/constraints-3.7.txt`. IOW, tests pass regardless if it is installed or not

testing/constraints-async-rest-3.7.txt
parthea: Having a separate file will allow us to test the minimum versions of the dependencies in the extra `async_rest`. If we add `aiohttp` to the existing constraints file then we won't have a configuration where `aiohttp` is not installed. For example, when we run `nox -s unit-*`, the `aiohttp` extra should not be installed. It should only be installed when we run `nox -s unit_w_async_rest_extra*`. 

It much easier to have this in a separate constraints file than hardcoded in `nox -s unit_w_async_rest_extra*`. We specifically want to test the minimum versions of dependencies in the extra and not the latest versions so simply having `pip install .[aiohttp]` in the nox session isn't sufficient.

vchudnov-g (commented):
One non-trivial comment about using the `extras` in `setup.py`

ohmayr:
> One non-trivial comment about using the `extras` in `setup.py`

What's the comment? I see that we're setting `extras` in `setup.py` [here](https://github.com/googleapis/python-api-core/blob/main/setup.py#L99)

noxfile.py
```diff
@@ -130,24 +131,40 @@ def default(session, install_grpc=True, prerelease=False, install_auth_aio=False
...
-            "-c",
-            f"{constraints_dir}/constraints-{session.python}.txt",
+            "-e", "." + ("[async_rest]" if install_async_rest else ""), "--no-deps"
         )
+    else:
+        constraints_file = (
+            f"{constraints_dir}/constraints-{constraints_type}{session.python}.txt"
+        )
+        # fall back to standard constraints file
+        if not pathlib.Path(constraints_file).exists():
```
ohmayr, line 149: @vchudnov-g shall we get rid of this section? Given that we've added a new constraints file, I think this is extra work and probably not needed. WDYT?

testing/constraints-async-rest-3.7.txt
vchudnov-g: _(my response below copied from [a comment on gapic-generator PR #2195](https://github.com/googleapis/gapic-generator-python/pull/2195#discussion_r1776152166)) on the same issue)_

I'm not sure this is true. The `nox` documentation [says](https://nox.thea.codes/en/stable/config.html) "[session.install] invokes pip to install packages inside of the session’s virtualenv." We use `session.install` like [this](https://github.com/googleapis/gapic-generator-python/blob/ab19c5a09b6ba1222cae4a207c7e4a4da4f83e21/gapic/templates/noxfile.py.j2#L57), specifying `-c CONSTRAINTS_FILE`. The `-c` constraints files option for `pip` is [documented](https://pip.pypa.io/en/latest/user_guide/#constraints-files) as saying (emphasis added) "Constraints files are requirements files that only control which version of a requirement is installed, _not whether it is installed or not_" .

So in other words, removing `async_rest` from the constraints will not do what we want. It doesn't matter whether we have it there or not, as long as we make sure that we do not _install_ the extras. So I would argue it's better to *not* split the constraints files, but to have one test suite that installs the `async_rest` extra and one test suite that doesn't, both with the same constraints file. 

Does this make sense? LMK if I've misunderstood something.

noxfile.py
ohmayr: As discussed offline, the constraints file will only enforce the versions but not install the requirements, so this is needed.

ohmayr (approved):

noxfile.py
```diff
@@ -130,24 +131,40 @@ def default(session, install_grpc=True, prerelease=False, install_auth_aio=False
...
         )
         # This *must* be the last install command to get the package from source.
-        session.install("-e", ".", "--no-deps")
-    else:
         session.install(
-            "-e",
-            ".[grpc]" if install_grpc else ".",
-            "-c",
-            f"{constraints_dir}/constraints-{session.python}.txt",
+            "-e", "." + ("[async_rest]" if install_async_rest else ""), "--no-deps"
```
vchudnov-g, line 142: This was a pre-existing issue: at the top of the function we require that `install_grpc` be set if `prerelease` is set. But here we don't install `grpc`. Is that an oversight? It seems to me we probably should install it here.

noxfile.py
```diff
@@ -130,24 +131,40 @@ def default(session, install_grpc=True, prerelease=False, install_auth_aio=False
...
         )
+    else:
+        constraints_file = (
+            f"{constraints_dir}/constraints-{constraints_type}{session.python}.txt"
+        )
+        # fall back to standard constraints file
+        if not pathlib.Path(constraints_file).exists():
+            constraints_file = f"{constraints_dir}/constraints-{session.python}.txt"
```
vchudnov-g, line 150: Add a reminder  at the top of the else that the constraints file does not determine *whether* the listed imports are imported.

vchudnov-g (approved):

noxfile.py
```diff
@@ -110,7 +111,7 @@ def install_prerelease_dependencies(session, constraints_path):
         session.install(*other_deps)
 
 
-def default(session, install_grpc=True, prerelease=False, install_auth_aio=False):
+def default(session, install_grpc=True, prerelease=False, install_async_rest=False):
```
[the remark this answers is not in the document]
ohmayr: addressed.

noxfile.py
ohmayr: addressed.

noxfile.py
ohmayr: addressed.

PR #676, opened 2024-07-06
fix: remove use of old external mock

a-detiste:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue:  there were various attempts before that got stuck by lack of CLA
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

PR #713, opened 2024-10-02
fix: Switch to unittest.mock from mock

### LINKED ISSUE #377: use unittest.mock instead of mock
Rather a question, do you plan to migrate mock usage in tests to built-in unittest.mock instead?

s-t-e-v-e-n-k:
Now that the minimum supported version of Python is 3.7, we can stop using the external mock requirement, and import it from unittest. I have also attempted to keep imports ordered.

Fixes #377
Closes #676

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -124,6 +124,7 @@ def default(session, install_grpc=True, prerelease=False, install_async_rest=Fal
 
     session.install(
         "dataclasses",
+        "mock; python_version=='3.7'",
         "pytest",
         "pytest-cov",
         "pytest-xdist",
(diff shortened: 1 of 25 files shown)


ohmayr:
@s-t-e-v-e-n-k, general comment: `AsyncMock` isn't part of the stdlib i.e. `unitest.mock` for Python 3.7. which we're using in our test code.

s-t-e-v-e-n-k:
Ahh, indeed, AsyncMock was added in 3.8. I'll modify this accordingly.

ohmayr (approved):

PR #733, opened 2024-10-09
fix: store prefetched exceptions

### LINKED ISSUE #731: Streams with prefecth_first_result do not convert grpc errors
one task of grpc_helpers is to wrap grpc channels in `error_remapped_callable`, converting grpc errors into `api_core.exception` types. But this error remapping is skipped when processing a streaming rpc with [prefetch_first](https://github.com/googleapis/python-api-core/blob/2100f7869d9dc4b75c594c95c51e4db8c392834f/google/api_core/grpc_helpers.py#L169) enabled.

When prefetch_first is enabled, the StreamingResponseIterator [calls next as part of its init](https://github.com/googleapis/python-api-core/blob/2100f7869d9dc4b75c594c95c51e4db8c392834f/google/api_core/grpc_helpers.py#L92). But if there is an error as part of the next call, the grpc error is raised immediately, without being wrapped

Because the error is not wrapped, and it is raised outside the usual call loop, this will cause retryable errors to not be retried properly

---

To solve this, we should probably [add a case to this try/catch](https://github.com/googleapis/python-api-core/blob/2100f7869d9dc4b75c594c95c51e4db8c392834f/google/api_core/grpc_helpers.py#L92):
```
       except grpc.RpcError as exc:
            self._stored_first_result = exc
```

And then raise it as part of \_\_next\_\_ [if present here](https://github.com/googleapis/python-api-core/blob/2100f7869d9dc4b75c594c95c51e4db8c392834f/google/api_core/grpc_helpers.py#L112)

related: b/368027810

daniel-sanche:
Resolves issues with pre-fetching streams not being retried when they should be

More context in attached bug

Fixes: https://github.com/googleapis/python-api-core/issues/731

Will add tests if there are no objections to the change

### DIFF
--- a/google/api_core/grpc_helpers.py
+++ b/google/api_core/grpc_helpers.py
@@ -99,7 +99,7 @@ class _StreamingResponseIterator(Generic[P], grpc.Call):
             pass
         except grpc.RpcError as exc:
             # If the pre-fetch fails, store exception to be raised on next() call.
-            self._stored_first_result = exc
+            self._stored_first_error = exc
 
     def __iter__(self) -> Iterator[P]:
         """This iterator is also an iterable that returns itself."""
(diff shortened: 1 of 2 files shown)


parthea:
daniel-sanche@ SG. Please assign back once tests are added and presubmits are green

google/api_core/grpc_helpers.py
```diff
@@ -113,6 +116,10 @@ def __next__(self) -> P:
                 result = self._stored_first_result
                 del self._stored_first_result
                 return result
+            elif hasattr(self, "_stored_first_error"):
+                exc = self._stored_first_error
+                del self._stored_first_error
+                raise exc
```
daniel-sanche, line 122: I'm considering whether it's worth refactoring both these variables, so we don't have to do two `hasattr` lookups for each stream result. For now I'm just following the existing code though

PR #740, opened 2024-10-21
fix: require proto-plus >= 1.25.0 for Python 3.13

### LINKED ISSUE #737: `proto-plus` >= 1.25 needed for python 3.13
The `setup.py` file should be updated to require `proto-plus` >= 1.25 for `python_version >=3.13` due to https://github.com/googleapis/proto-plus-python/issues/490

See release notes here: https://github.com/googleapis/proto-plus-python/releases/tag/v1.25.0

Add `"proto-plus >= 1.25.0, <2.0.0dev; python_version>='3.13'",` after the following line

https://github.com/googleapis/python-api-core/blob/46b3d3abaa1bae28e9d788d7c3006224cd6f74d5/setup.py#L34

parthea:
Fixes #737 🦕


parthea:
Wait for https://github.com/googleapis/proto-plus-python/issues/495

setup.py
```diff
@@ -31,7 +31,8 @@
 dependencies = [
     "googleapis-common-protos >= 1.56.2, < 2.0.dev0",
     "protobuf>=3.19.5,<6.0.0.dev0,!=3.20.0,!=3.20.1,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
-    "proto-plus >= 1.22.3, <2.0.0dev",
+    "proto-plus >= 1.22.3, < 2.0.0dev",
+    "proto-plus >= 1.25.0, < 2.0.0dev; python_version >= '3.13'",
```
ohmayr, line 35: How does this work?

Let's say if I have python 3.13 installed in my environment, will it install a `proto-plus` version first that satisfies this requirement `"proto-plus >= 1.22.3, < 2.0.0dev",` (since it doesn't specify any Python version) and then update the version when it gets to this statement `"proto-plus >= 1.25.0, < 2.0.0dev; python_version >= '3.13'",`? If that's the case, we should update line 34 to be `"proto-plus >= 1.22.3, < 2.0.0dev"; python_version < '3.13'"`.

ohmayr (approved):

setup.py
ohmayr: Discussed this offline. No changes required here.

PR #746, opened 2024-10-25
chore: update templated files

parthea:
Update the post processor image to the version which includes the fix from https://github.com/googleapis/synthtool/pull/2028


ohmayr (approved):

ohmayr:
Can you explain why we're moving from staging to dev?

parthea:
The specific change to move from staging to dev is part of https://github.com/googleapis/synthtool/pull/2026. Based on internal issue b/374954111, the staging environment was deleted in favor of a dev environment.

PR #754, opened 2024-11-04
feat: add automatic logging config to support debug logging

ohmayr:
Implement automatic logging config to allow users to enable debug logging via an environment variable.

Fixes b/381099872


google/api_core/client_logging.py
```diff
@@ -0,0 +1,73 @@
+import logging
+import json
+import re
+import os
+
+LOGGING_INITIALIZED = False
```
vchudnov-g, line 6: Should we have an underscore prefix?

google/api_core/client_logging.py
```diff
@@ -0,0 +1,73 @@
...
+
+# TODO(<add-link>): Update Request / Response messages.
+REQUEST_MESSAGE = "Sending request ..."
+RESPONSE_MESSAGE = "Receiving response ..."
+
+# TODO(<add-link>): Update this list to support additional logging fields
+_recognized_logging_fields = ["httpRequest", "rpcName", "serviceName"] # Additional fields to be Logged.
+
+def logger_configured(logger):
+  return logger.hasHandlers() or logger.level != logging.NOTSET
```
vchudnov-g, line 16: For completeness, I would add `or logger.propagate = False`

google/api_core/client_logging.py
```diff
@@ -0,0 +1,73 @@
...
+   if LOGGING_INITIALIZED:
+     return
+   scopes = os.getenv("GOOGLE_SDK_PYTHON_LOGGING_SCOPE")
+   setup_logging(scopes)
+   LOGGING_INITIALIZED = True
+
+def parse_logging_scopes(scopes):
+  if not scopes:
+     return []
+  # TODO(<add-link>): check if the namespace is a valid namespace.
```
vchudnov-g, line 29: My understanding was that we would check for valid scopes in the current work, and the follow-up item was to consider issuing a warning in case of an invalid scope.

google/api_core/client_logging.py
```diff
@@ -0,0 +1,73 @@
...
+  # TODO(<add-link>): parse a list of namespaces. Current flow expects a single string for now.
+  namespaces = [scopes]
+  return namespaces
+
+def default_settings(logger):
```
vchudnov-g, line 34: nit: I suggest naming functions as verbs. How about `configure_defaults`?

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,17 @@
+import logging
+import pytest
+
+from google.api_core.client_logging import BaseLogger
```
vchudnov-g, line 4: Where is `BaseLogger` defined?

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,17 @@
...
+
```
vchudnov-g, line 5: I suggest adding some cases testing `setup_logging` with no scopes, the `google` scope, a `google.foo` scope, and an `random.bar` scope, and then testing the expected behavior: warnings, propagation, handler+formatter. WDYT?

google/api_core/client_logging.py
ohmayr: addressed.

google/api_core/client_logging.py
```diff
@@ -0,0 +1,76 @@
+import logging
+import json
+import re
+import os
+
+_LOGGING_INITIALIZED = False
+_BASE_LOGGER_NAME = "google"
+# TODO(https://github.com/googleapis/python-api-core/issues/760): Update Request / Response messages.
+REQUEST_MESSAGE = "Sending request ..."
+RESPONSE_MESSAGE = "Receiving response ..."
```
vchudnov-g, line 10:
```suggestion
REQUEST_MESSAGE = "Sending request "
RESPONSE_MESSAGE = "Receiving response "
```
(space at the end intentional if we want to later append the RPC name)

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,62 @@
+import logging
+import pytest
+
+from google.api_core.client_logging import setup_logging
```
vchudnov-g, line 4: Another testcase would be this sequence:
1. Mock so that `os.getenv` returns `google.foo`
2. `initialize_logging()` 
3. check configs are what we expect for `google` and `google.foo`
4. Change the logger configs for  BOTH `google.foo` and `google`
5. Call `initialize_logging()` again
6. Check that the changes you made in step 4 are still there

google/api_core/client_logging.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for namespace in logger_names:
+      # This will either create a module level logger or get the reference of the base logger instantiated above.
+      logger = logging.getLogger(namespace)
+
+      # Configure default settings.
+      configure_defaults(logger)
+
+    # disable log propagation at base logger level to the root logger only if a base logger is not already configured via code changes.
+    # Maybe we do this at the end?
```
vchudnov-g, line 58:
```suggestion
```

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,62 @@
...
+    assert base_logger.handlers == []
+    assert base_logger.propagate == False
+    assert base_logger.level == logging.NOTSET
+
+    # TODO: update test once we add logic to ignore an incorrect scope.
+    logger = logging.getLogger("foo")
+    assert isinstance(logger.handlers[0], logging.StreamHandler)
+    assert logger.propagate == False
+    assert logger.level == logging.DEBUG
+
```
vchudnov-g, line 61:
```suggestion
reset_logger("foo")

```

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,62 @@
...
+    assert base_logger.propagate == False
+    assert base_logger.level == logging.NOTSET
+
+    module_logger = logging.getLogger("google.foo")
+    assert isinstance(module_logger.handlers[0], logging.StreamHandler)
+    assert module_logger.propagate == False
+    assert module_logger.level == logging.DEBUG
+
+
+    reset_logger("google")
```
vchudnov-g, line 46:
```suggestion
    reset_logger("google.foo")
    reset_logger("google")    
```

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_setup_logging_w_incorrect_scope():
+    setup_logging("foo")
+    
+    base_logger = logging.getLogger("google")
+    assert base_logger.handlers == []
+    assert base_logger.propagate == False
+    assert base_logger.level == logging.NOTSET
+
+    # TODO: update test once we add logic to ignore an incorrect scope.
```
vchudnov-g, line 56:
```suggestion
    # TODO(https://github.com/googleapis/python-api-core/issues/759): update test once we add logic to ignore an incorrect scope.
```

google/api_core/client_logging.py
```diff
@@ -0,0 +1,89 @@
+import logging
+import json
+import re
+import os
+
+_LOGGING_INITIALIZED = False
+_BASE_LOGGER_NAME = "google"
+# TODO(https://github.com/googleapis/python-api-core/issues/760): Update Request / Response messages.
+REQUEST_MESSAGE = "Sending request "
+RESPONSE_MESSAGE = "Receiving response "
```
parthea, line 10: Consider keeping these in `gapic-generator-python` to avoid having to require newer versions of `google-api-core` when log events are added downstream

google/api_core/client_logging.py
```diff
@@ -0,0 +1,89 @@
...
+def initialize_logging():
+    global _LOGGING_INITIALIZED
+    if _LOGGING_INITIALIZED:
+        return
+    scopes = os.getenv("GOOGLE_SDK_PYTHON_LOGGING_SCOPE", "")
+    setup_logging(scopes)
+    _LOGGING_INITIALIZED = True
+
+
+def parse_logging_scopes(scopes):
```
parthea, line 37: Add typing

```suggestion
def parse_logging_scopes(scopes: Optional[str] = None):
```

tests/unit/test_client_logging.py
parthea: Re-opening as we should document/test the formatter that we expect in each of the test cases

google/api_core/client_logging.py
```diff
@@ -0,0 +1,89 @@
...
+        logger.addHandler(console_handler)
+
+
+def setup_logging(scopes=[]):
+
+    # only returns valid logger scopes (namespaces)
+    # this list has at most one element.
+    logger_names = parse_logging_scopes(scopes)
+
+    for namespace in logger_names:
```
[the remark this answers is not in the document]
vchudnov-g: If called with no scopes, the user didn't tell us we need to set up logging for everything, so I think all we should do stop propagation at the base logger, like we're doing now. In fact, i think we *need* to do this, because if there is no environment variable set and we set up structured logging, then the user will start getting logs as soon as they get to this version of the library, which is a no-no.

google/api_core/client_logging.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # disable log propagation at base logger level to the root logger only if a base logger is not already configured via code changes.
+    base_logger = logging.getLogger(_BASE_LOGGER_NAME)
+    if not logger_configured(base_logger):
+        base_logger.propagate = False
+
+
+class StructuredLogFormatter(logging.Formatter):
+    def format(self, record):
+        log_obj = {
```
parthea, line 78: It may be helpful to add the function name, file name and line number of where the log exists in the source code, similar to what exists in `google-cloud-logging`. 

For example, add `"line_number": record.lineno` and other helpful information from https://docs.python.org/3/library/logging.html#logrecord-objects

google/api_core/client_logging.py
vchudnov-g: You mean during development, if we modifiy the generator before we release `google-api-core`? We will need to require updated versions of the `google-api-core` when we release the logging-enabled GAPICs. Otherwise, log messages from the GAPICs will propagate to the root logger, which we don't want.

parthea: > We will need to require updated versions of the google-api-core when we release the logging-enabled GAPICs.

My initial thought is that we don't want to force users to update to the latest version of `google-api-core`. We can detect whether a newer version of `google-api-core` is installed, and avoid logging when there is an old version.


We don't want to force users onto newer versions of google-api-core every time that we want to define a new log. Ideally users can remain on old versions of `google-api-core` and only update as new functionality is needed (for example, if they choose to enable logging, they can only do so with a newer version of `google-api-core`).

tests/unit/test_client_logging.py
ohmayr: addressed.

tests/unit/test_client_logging.py
ohmayr: addressed.

tests/unit/test_client_logging.py
ohmayr: The formatter in each of the test case is structured format. I think testing anything from the default config is sufficient to confirm that they're applied. Let me know what you think!

google/api_core/client_logging.py
ohmayr: This can be addressed as a follow up and is tracked here: https://github.com/googleapis/python-api-core/issues/761.

google/api_core/client_logging.py
ohmayr: addressed.

google/api_core/client_logging.py
ohmayr: I think the setup defined here and any new log events that we add in our gapic should be guarded.

tests/unit/test_client_logging.py
ohmayr: addressed.

google/api_core/client_logging.py
```diff
@@ -0,0 +1,85 @@
...
+    if not logger_configured(logger):
+        console_handler = logging.StreamHandler()
+        logger.setLevel("DEBUG")
+        logger.propagate = False
+        formatter = StructuredLogFormatter()
+        console_handler.setFormatter(formatter)
+        logger.addHandler(console_handler)
+
+
+def setup_logging(scopes=[]):
```
vchudnov-g, line 53:
```suggestion
def setup_logging(scopes=""):
```

We'll have this function take just one string, as would be read from the environment variable. `parse_logging_scopes` below will parse it into a list.

google/api_core/client_logging.py
vchudnov-g: SG. I would add a link to that issue in a comment.

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,103 @@
...
+    logger.propagate = True
+
+
+def test_setup_logging_w_no_scopes():
+    with mock.patch("google.api_core.client_logging._BASE_LOGGER_NAME", "foo"):
+        setup_logging()
+        base_logger = logging.getLogger("foo")
+        assert base_logger.handlers == []
+        assert not base_logger.propagate
+        assert base_logger.level == logging.NOTSET
```
vchudnov-g, line 23: We should be consistent with the other test cases as to whether we have the `getLogger` and the assertions within the `with` statement.

```suggestion
    with mock.patch("google.api_core.client_logging._BASE_LOGGER_NAME", "foo"):
        setup_logging()
    base_logger = logging.getLogger("foo")
    assert base_logger.handlers == []
    assert not base_logger.propagate
    assert base_logger.level == logging.NOTSET
```

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,103 @@
...
+    base_logger = logging.getLogger("foo")
+    assert base_logger.handlers == []
+    assert not base_logger.propagate
+    assert base_logger.level == logging.NOTSET
+
+    module_logger = logging.getLogger("foo.bar")
+    assert isinstance(module_logger.handlers[0], logging.StreamHandler)
+    assert not module_logger.propagate
+    assert module_logger.level == logging.DEBUG
+
```
vchudnov-g, line 91:
```suggestion

    ### Check that `initialize_logging()` is a no-op after the first time by verifying that user-set configs are not modified:
    
```

tests/unit/test_client_logging.py
```diff
@@ -0,0 +1,103 @@
...
+def reset_logger(scope):
+    logger = logging.getLogger(scope)
+    logger.handlers = []
+    logger.setLevel(logging.NOTSET)
+    logger.propagate = True
+
+
+def test_setup_logging_w_no_scopes():
+    with mock.patch("google.api_core.client_logging._BASE_LOGGER_NAME", "foo"):
```
vchudnov-g, line 18: nit (not blocking) : For readability and a more informative mock, consider s/`foo`/`foogle`/g or s/`foo`/`test_base_logger`/g everywhere in this file. We can also do this as a follow-up.

(Again, not a blocker)

tests/unit/test_client_logging.py
vchudnov-g: @ohmayr : OK, that sounds fine. Separately, though, it would be useful to have a test for the formatter, where we call `StructuredLogFormatter.format(sample_data)` directly and verify we have the JSON output as expected. It would be a sort of golden test for the formatter (field names and values populated correctly). We can do it as a follow-up. (Maybe add a TODO with an issue link in this file?)

google/api_core/client_logging.py
```diff
@@ -0,0 +1,85 @@
...
+]  # Additional fields to be Logged.
+
+
+def logger_configured(logger):
+    return (
+        logger.handlers != [] or logger.level != logging.NOTSET or not logger.propagate
+    )
+
+
+def initialize_logging():
```
vchudnov-g, line 24: We should add documentation to these various functions. We can do it as a follow-up. (File a bug?)

tests/unit/test_client_logging.py
ohmayr: addressed.

google/api_core/client_logging.py
vchudnov-g: > We don't want to force users onto newer versions of google-api-core every time that we want to define a new log. Ideally users can remain on old versions of google-api-core and only update as new functionality is needed (for example, if they choose to enable logging, they can only do so with a newer version of google-api-core).

I don't follow. I'm not suggesting users update `python-api-core` "every time that we want to define a new log". I'm suggesting they update ONCE, with essentially the functionality in this PR. New GAPICs will depend on this functionality regardless of where/when we add the log statements.

> any new log events that we add in our gapic should be guarded.

If we have the GAPICs depend on this version of `core`, then they won't be needed to be guarded, which makes the generator changes more sustainable. (Regardless, any guarding we do should be well encapsulated so logging an event remains a one liner.)

parthea: What happens when we want to add another message like `REQUEST_MESSAGE = "Sending request "`, for example, `REQUEST_MESSAGE_METADATA = "Initial Metadata: "`? Users will need to upgrade `google-api-core` to be able to see those logs.

vchudnov-g (approved):

parthea (approved):

google/api_core/client_logging.py
```diff
@@ -0,0 +1,141 @@
...
+        console_handler = logging.StreamHandler()
+        logger.setLevel("DEBUG")
+        logger.propagate = False
+        formatter = StructuredLogFormatter()
+        console_handler.setFormatter(formatter)
+        logger.addHandler(console_handler)
+
+
+# TODO(https://github.com/googleapis/python-api-core/issues/763): Expand documentation.
+def setup_logging(scopes=""):
```
parthea, line 92: Add type hints
```suggestion
def setup_logging(scopes: str =""):
```

google/api_core/client_logging.py
```diff
@@ -0,0 +1,141 @@
...
+# TODO(https://github.com/googleapis/python-api-core/issues/761): Update this list to support additional logging fields.
+_recognized_logging_fields = [
+    "httpRequest",
+    "rpcName",
+    "serviceName",
+]  # Additional fields to be Logged.
+
+
+# TODO(https://github.com/googleapis/python-api-core/issues/763): Expand documentation.
+def logger_configured(logger):
```
parthea, line 21:
```suggestion
def logger_configured(logger) -> bool:
```

We should also include the type of `logger`

google/api_core/client_logging.py
```diff
@@ -0,0 +1,141 @@
...
+    base_logger = logging.getLogger(_BASE_LOGGER_NAME)
+    if not logger_configured(base_logger):
+        base_logger.propagate = False
+
+
+# TODO(https://github.com/googleapis/python-api-core/issues/763): Add documentation.
+class StructuredLogFormatter(logging.Formatter):
+    # TODO(https://github.com/googleapis/python-api-core/issues/761): ensure that additional fields such as
+    # function name, file name, and line no. appear in a log output.
+    def format(self, record):
```
parthea, line 129: Include the type hint for `record`

google/api_core/client_logging.py
```diff
@@ -0,0 +1,141 @@
...
+
+from typing import List, Optional
+
+_LOGGING_INITIALIZED = False
+_BASE_LOGGER_NAME = "google"
+
+# Fields to be included in the StructuredLogFormatter.
+#
+# TODO(https://github.com/googleapis/python-api-core/issues/761): Update this list to support additional logging fields.
+_recognized_logging_fields = [
```
parthea, line 13: One thing to note is that users will need to update `google-api-core` to get any additional logging fields in the future.

google/api_core/client_logging.py
```diff
@@ -0,0 +1,141 @@
...
+    "httpRequest",
+    "rpcName",
+    "serviceName",
```
parthea, line 16: I just want to confirm that we are ok with deviating from the standard snake case that's typically used

https://peps.python.org/pep-0008/#function-and-variable-names

eg.

```suggestion
    "http_request",
    "rpc_name",
    "service_name",
```

google/api_core/client_logging.py
```diff
@@ -0,0 +1,141 @@
...
+]  # Additional fields to be Logged.
```
parthea, line 17:
```suggestion
    "credentialsType",
    "credentialsScope",
    "useSelfSignedJWT",
    "universeDomain",
]  # Additional fields to be Logged.
```
Add credentials related fields so that we can include information similar to the ruby library

```
I, [2024-09-30T16:52:15.809067 #1917313]  INFO -- google-cloud-secret_manager-v1: Created client for google.cloud.secretmanager.v1.SecretManagerService -- {"clientId":3220,"system":"google-cloud-secret_manager-v1","serviceName":"google.cloud.secretmanager.v1.SecretManagerService","credentialsId":3260,"credentialsType":"Google::Auth::ServiceAccountCredentials","credentialsScope":["https://www.googleapis.com/auth/cloud-platform"],"useSelfSignedJWT":false,"universeDomain":"googleapis.com"}
```

parthea (approved):

google/api_core/client_logging.py
ohmayr: Yes it is fine and in fact a requirement in some cases to have camel case names.

google/api_core/client_logging.py
ohmayr: This is true. Any fields that we add in the future will require users to update `google-api-core`.

PR #766, opened 2024-12-05
feat: update recognized logging fields

ohmayr:


google/api_core/client_logging.py
```diff
@@ -12,10 +12,11 @@
 # TODO(https://github.com/googleapis/python-api-core/issues/761): Update this list to support additional logging fields.
 _recognized_logging_fields = [
     "httpRequest",
-    "rpcName",
+    "httpRequestMethod",
+    "httpRequestUrl" "rpcName",
```
vchudnov-g, line 16: As per https://cloud.google.com/logging/docs/structured-logging#structured_logging_special_fields, `httpRequest` should be of type `HttpRequest`, [which](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest) includes the fields `requestMethod` and `requestURL` internally.


```suggestion
    "rpcName",
```

google/api_core/client_logging.py
```diff
@@ -15,12 +15,13 @@
     "rpcName",
     "serviceName",
     "credentialsType",
-    "credentialInfo",
+    "credentialsInfo",
     "universeDomain",
     "request",
     "response",
     "metadata",
```
vchudnov-g, line 22: Per the parent design doc, we should still have `rpcName`, right? (We're using it in the pending PR in the generator)

ohmayr: Yes we should. I see it here it here in the list.

vchudnov-g (approved):

PR #770, opened 2024-12-16
fix: memory leak in bidi classes

### LINKED ISSUE #769: Memory leak in bidi resources
Reported in python-firestore: https://github.com/googleapis/python-firestore/issues/985

It looks like some of the resources in BackgroundConsumer and ResumableBidiRpc are not being cleaned up properly on close, and preventing garbage collection. I ran the attached script, and can confirm I see the same thing

I opened https://github.com/googleapis/python-api-core/pull/770 to address the issue, but maybe someone with more familiarity with the classes can take a look as well

daniel-sanche:
Fixes https://github.com/googleapis/python-api-core/issues/769


google/api_core/bidi.py
```diff
@@ -717,6 +719,7 @@ def stop(self):
                     _LOGGER.warning("Background thread did not exit.")
 
             self._thread = None
+            self._on_response = None
```
daniel-sanche, line 722: This assumes you can't restart an instance after it has been stopped. The test cases and docstrings point to that being the case, but it's not completely clear

parthea: I think this is a reasonable assumption given that resources should be cleaned up when `close` is called. Regardless, we'll publish a pre-release to get feedback on this change before publishing a normal release.

parthea (approved):

parthea:
Thanks @daniel-sanche !

PR #793, opened 2025-02-14
feat: support dynamic retry backoff values

### LINKED ISSUE #792: Retry should generate sleep time after attempt
Currently our retry logic [pulls the next sleep time from the generator](https://github.com/googleapis/python-api-core/blob/7fbd5fda207d856b5835d0e0166df52e4819522a/google/api_core/retry/retry_streaming_async.py#L114) before starting an rpc. This allows us to [immediately fail with a ValueError](https://github.com/googleapis/python-api-core/blob/7fbd5fda207d856b5835d0e0166df52e4819522a/google/api_core/retry/retry_streaming_async.py#L192) if the generator stops providing backoff values

The downside of this approach is that the backoff is locked in before the rpc is attempted.

Bigtable [has a feature](https://github.com/googleapis/python-bigtable/issues/1064) where the server can provide a backoff value to the client for the next attempt. The design of Python's retry classes is flexible enough to allow this, except for the fact that the next backoff value is locked in early.

To support this feature, I suggest that we only call the backoff generator when we are ready for the next sleep.

This change would break some existing tests, because faulty backoff generators now won't be caught until after the first rpc has failed. But I don't believe real code should be impacted

daniel-sanche:
This PR moves backoff generation until after an rpc is complete and on_error callbacks have been called. This gives the client a chance to change what backoff value is used based on the error details.

This feature is required for Bigtable, where the backend can send retry_info packets to inform the client on how long to wait before a retry

Fixes https://github.com/googleapis/python-api-core/issues/792


google/api_core/retry/retry_streaming.py
```diff
@@ -107,8 +107,9 @@ def retry_target_stream(
         time.monotonic() + timeout if timeout is not None else None
     )
     error_list: list[Exception] = []
+    sleep_iter = iter(sleep_generator)
 
-    for sleep in sleep_generator:
+    while True:
```
parthea, line 112: Please could you file a follow up bug to refactor this code to add another exit condition? It's not ideal to use `while True` because there could be a scenario where the code is stuck in the while loop forever. Please could you add a comment clarifying the reason this loop will never be stuck indefinitely? If there is any doubt, we may need to block this PR until another exit condition is added. Otherwise, this is a non-blocking comment. Applies throughout.

parthea (commented):
Local testing looks good. Test results in https://github.com/googleapis/gapic-generator-python/pull/2359 look good. I'm also waiting for test results in cl/736133031. Once the tests in cl/736133031 pass and we are confident that there won't be an infinite loop ( due to use of `while True` without another exit condition), we can move forward with this PR.

google/api_core/retry/retry_streaming.py
daniel-sanche: The logic should be the same as before. We've always relied on the wrapped function to tell us when to terminate (either by returning, or raise an exception)

When writing the code though, I did consider that now might be a good time to add a [max_attempts argument](https://github.com/googleapis/python-api-core/issues/535) as well, which would give us an exit condition here. Do you think I should consider implementing that as well? Or maybe adding a TODO referencing that it will come later?

parthea: If there is no concern that there will be an infinite loop (eg. same behavior as before), a TODO comment will be sufficient. In terms of functional testing, I haven't been able to create a problem locally.

daniel-sanche: sounds good, just added the comment

google/api_core/retry/retry_base.py
```diff
@@ -212,6 +214,12 @@ def _retry_error_helper(
         raise final_exc from source_exc
     if on_error_fn is not None:
         on_error_fn(exc)
+    # next_sleep is fetched after the on_error callback, to allow clients
+    # to update sleep_iterator values dynamically in response to errors
+    try:
+        next_sleep = next(sleep_iterator)
```
parthea, line 220: Please could you add a regression test which requires this new functionality? If this functionality is removed, the tests in this PR still pass.

daniel-sanche: FWIW, there shouldn't be new functionality, this is just moving what used to be [at the top of the loop](https://github.com/googleapis/python-api-core/blob/a5648faf31c30158d59a3127ea37f7c9911ff765/google/api_core/retry/retry_streaming.py#L111) to the bottom

But yeah, I can probably come up with a test that tests the new sequencing of events if you think that would be helpful to have

parthea: The specific ask is for a test that fails at `main` without the code changes in this PR to avoid regressions after this PR is merged.

daniel-sanche: Yeah sounds good, I'll add a test for the sequencing (it might slip to next week though)

parthea:
I'm going to move this to draft until a regression test is added. Please mark it as ready to review when this PR is ready.

parthea (approved):

PR #794, opened 2025-02-14
feat: add client debug logging support for streaming gRPC/REST calls

parthea:
Towards b/383403001

### DIFF
--- a/google/api_core/_rest_streaming_base.py
+++ b/google/api_core/_rest_streaming_base.py
@@ -15,6 +15,7 @@
 """Helpers for server-side streaming in REST."""
 
 from collections import deque
+import logging
 import string
 from typing import Deque, Union
 import types
@@ -22,6 +23,8 @@ import types
 import proto
 import google.protobuf.message
 from google.protobuf.json_format import Parse
+
+_LOGGER = logging.getLogger(__name__)
 
 
 class BaseResponseIterator:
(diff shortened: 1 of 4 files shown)


google/api_core/_rest_streaming_base.py
```diff
@@ -97,19 +100,38 @@ def _process_chunk(self, chunk: str):
                 self._obj += char
             self._escape_next = not self._escape_next if char == "\\" else False
 
+    def _log_response_payload(self, response_payload: str):  # pragma: NO COVER
```
ohmayr, line 103: Why do we need to skip coverage here?

google/api_core/_rest_streaming_base.py
```diff
@@ -97,19 +100,38 @@ def _process_chunk(self, chunk: str):
...
+        rest_response = {
+            "payload": response_payload,
+            "status": "OK",
+        }
+        _LOGGER.debug(
+            "Received response via REST stream",
+            extra={
+                "response": rest_response,
```
ohmayr, line 111: should we be using `httpResponse` here instead for structured logs?

google/api_core/_rest_streaming_base.py
```diff
@@ -97,19 +100,38 @@ def _process_chunk(self, chunk: str):
...
+                    response_payload, ignore_unknown_fields=True
                 )
 
             return grab
         elif issubclass(self._response_message_cls, google.protobuf.message.Message):
 
             def grab(this):
-                return Parse(this._ready_objs.popleft(), this._response_message_cls())
+                response_payload = this._ready_objs.popleft()
+                if logging_enabled:  # pragma: NO COVER
```
ohmayr, line 132: Can we add coverage for this?

google/api_core/grpc_helpers.py
```diff
@@ -112,8 +117,29 @@ def __next__(self) -> P:
...
+                    )
+                grpc_response = {
+                    "payload": response_payload,
+                    "status": "OK",
+                }
+                _LOGGER.debug(
+                    f"Received response of type {type(result)} via gRPC stream",
+                    extra={
+                        "response": grpc_response,
+                    },
```
ohmayr, line 140: Can this be added as a helper function and re-used in `grpc_helpers_async`?

google/api_core/grpc_helpers_async.py
```diff
@@ -94,7 +100,28 @@ def __init__(self):
 
     async def read(self) -> P:
         try:
-            return await self._call.read()
+            result = await self._call.read()
+            logging_enabled = _LOGGER.isEnabledFor(logging.DEBUG)
+            if logging_enabled:  # pragma: NO COVER
+                if isinstance(result, proto.Message):
```
ohmayr, line 106: Can you link 

```
{# TODO(https://github.com/googleapis/gapic-generator-python/issues/2293): Investigate if we can improve this logic or wait for next gen protobuf. #}
```

ohmayr: Also, can the parsing logic be added as a helper function and re-used in the gapic? We can file a TODO for the latter.

vchudnov-g: +1 

This ties into the issue I still want to get back to b/382299158. Maybe we reference that issue instead of creating a new one, and in that issue we reference these changes so we have easy access.

google/api_core/grpc_helpers_async.py
```diff
@@ -94,7 +100,28 @@ def __init__(self):
 
     async def read(self) -> P:
         try:
-            return await self._call.read()
+            result = await self._call.read()
```
vchudnov-g, line 103: IIUC, this gRPC doing an actual read from the network, so logging it as such below makes sense.


(just pointing this out to contrast with my other comments about logging returning previously streamed messages to the user)

vchudnov-g:
Also, could you reference https://github.com/googleapis/gapic-generator-python/issues/2289 rather than the mirrored internal issue in this PR's description?

vchudnov-g:
@parthea  What's the status of this?

PR #851, opened 2025-10-28
fix: closes tailing streams in bidi classes.

### LINKED ISSUE #850: Tailing streams on closing a bidi stream.
The `close()` method of the bidi module, resets the fields only when the self.call field is not None. 

```
    def close(self):
        """Closes the stream."""
        if self.call is None:
            return

        self._request_queue.put(None)
        self.call.cancel()
        ....
```

This is problematic, when an exception is raised in the `open()` and we do not set the self.call. This scenario leads to the stream not getting closed properly(half closed state from server). And we are able to observe these tailing streams(which are causing further issues for us) incase of reopens and other retries. Also observed that these tailing streams get closed automatically after some time(15-20 mins).

#### Environment details
Using gcloud installation.

  - OS type and version: Ubuntu 22.04.5 LTS, Mac OS Sequoia 15.7.1. Not tested on other OS.
  - Python version: Bundled python which comes with gcloud.
  - `google-api-core` version: Using gcloud installation version with 545.0.0

Sharing further information internally(@pujawadare googler),  as some information may be sensitive.

pujawadare:
Always put `None` into the request queue when closing a bidi stream. This ensures that the request queue is always signaled as closed, even if the underlying gRPC call object is not yet available.

Fixes #850  🦕

### DIFF
--- a/google/api_core/bidi.py
+++ b/google/api_core/bidi.py
@@ -281,11 +281,11 @@ class BidiRpc(BidiRpcBase):
 
     def close(self):
         """Closes the stream."""
-        if self.call is None:
-            return
+        if self.call is not None:
+            self.call.cancel()
 
+        # Put None in request queue to signal termination.
         self._request_queue.put(None)
-        self.call.cancel()
         self._request_generator = None
         self._initial_request = None
         self._callbacks = []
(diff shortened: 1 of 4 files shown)


daniel-sanche:
I think we'll need a similar change on the async side as well: https://github.com/chandra-siri/python-api-core/blob/82cea4f767510be2002fcc5807a770b99a2906c1/google/api_core/bidi_async.py#L198

pujawadare (commented):
Also included async bidi classes in the fix.

daniel-sanche (dismissed):

daniel-sanche (approved):

chalmerlowe (approved):
LGTM

Issue #857, opened 2025-10-31
`An error occurred: module 'importlib.metadata' has no attribute 'packages_distributions'`


ReubenFrankel:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/python-api-core/issues
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS type and version: Ubuntu 22.04.5 LTS
  - Python version: Python 3.9.21
  - pip version: 24.3.1
  - `google-api-core` version: 2.28.1

#### Steps to reproduce

  1. Import `google.api_core` in Python REPL
  2. Observed error printed to stdout

#### Code example

```console
$ .meltano/extractors/tap-bigquery/venv/bin/python
Python 3.9.21 (main, Dec  3 2024, 17:50:13) 
[GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
Ctrl click to launch VS Code Native REPL
>>> import google.api_core
An error occurred: module 'importlib.metadata' has no attribute 'packages_distributions'
/home/reuben/meltano/bigquery-test/.meltano/extractors/tap-bigquery/venv/lib/python3.9/site-packages/google/api_core/_python_version_support.py:252: FutureWarning: You are using a Python version (3.9.21) past its end of life. Google will update google.api_core with critical bug fixes on a best-effort basis, but not with any other fixes or features. Please upgrade to the latest Python version, or at least Python 3.10, and then update google.api_core.
  warnings.warn(message, FutureWarning)
>>>
```

#### Stack trace
```
# example
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

---

This is happening because `importlib.metadata.packages_distributions` was added in Python 3.10, but the current version check implementation allows this to be called for 3.8 and above: https://github.com/googleapis/python-api-core/blob/c97b3a004044ebf8b35c2a7ba97409d7795e11b0/google/api_core/_python_version_support.py#L150-L176

In our case, this generally wouldn't be a problem if the error was logged to stderr rather than printed, as we use following a specification that is particularly sensitive to stdout pollution.

[...]

PR #858, opened 2025-10-31
fix: Log version check errors

ReubenFrankel:
- [ ] Ensure the tests and linter pass

Partially addresses #857 🦕

### DIFF
--- a/google/api_core/_python_version_support.py
+++ b/google/api_core/_python_version_support.py
@@ -172,11 +172,10 @@ else:
             if module_name in module_to_distributions:  # pragma: NO COVER
                 # The value is a list of distribution names, take the first one
                 return module_to_distributions[module_name][0]
-            else:
-                return None  # Module not found in the mapping
         except Exception as e:
             _LOGGER.error("An error occurred: %s", e)
-            return None
+
+        return None
 
 
 def _get_distribution_and_import_packages(import_package: str) -> Tuple[str, Any]:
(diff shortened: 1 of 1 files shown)


google/api_core/_python_version_support.py
```diff
@@ -168,11 +172,10 @@ def _get_pypi_package_name(module_name):
             if module_name in module_to_distributions:  # pragma: NO COVER
                 # The value is a list of distribution names, take the first one
                 return module_to_distributions[module_name][0]
-            else:
-                return None  # Module not found in the mapping
         except Exception as e:
-            print(f"An error occurred: {e}")
-            return None
+            _LOGGER.error("An error occurred: %s", e)
```
chalmerlowe: This is an informational announcement and is considered no-op (i.e. we swallow the `Exception` and simply return `None` which is an allowed response. I recommend that we change the logger type to `.info()` and add a smidge more context in the message.)

```suggestion
            _LOGGER.info(f"An error occurred while determining PyPI package name for {module_name}: {e}")
```

ReubenFrankel:
> If you have a strong objection to the suggestion, I am happy to hear your reasoning.

No objections, I kept the message the same only because that was what was there before and the intention of the PR was to move from `print` to logging.

chalmerlowe (dismissed):
LGTM.

parthea (approved):

Linchin (approved):

PR #874, opened 2025-11-26
tests: refactor lint nox sessions

parthea:
This change is needed as part of b/463296248. Instead of having 2 different presubmits that run lint and mypy checks, we now have these 2 patterns tested under a single nox session. This follows the pattern that we have in google-cloud-python where there is a single lint nox session.

The `pytype` nox session was not running as a presubmit so I've removed it. `pytype` has been deprecated as per [this note](https://github.com/google/pytype?tab=readme-ov-file#an-update-on-pytype).


noxfile.py
```diff
@@ -265,27 +276,6 @@ def lint_setup_py(session):
...
-    """Run type-checking."""
-    session.install(".[grpc,async_rest]", "mypy")
-    session.install(
-        "types-setuptools",
-        "types-requests",
-        "types-protobuf",
-        "types-dataclasses",
-        "types-mock; python_version=='3.7'",
-    )
-    session.run("mypy", "google", "tests")
```
daniel-sanche: I do think there's value in being able to run mypy on its own sometimes. 

It could be annoying to have to fix all other style issues, before even seeing if the types are valid

parthea: Ack. I'll keep mypy as a separate presubmit

